PR #1, opened 2022-09-15
Remove Rust nightly dependency

BradLugo:


src/lib.rs
```diff
@@ -28,11 +28,6 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
-// Issue tracking: https://github.com/rust-lang/rust/issues/85410
-// Here we need to build an abstract socket from a SocketAddr until
-// tokio supports abstract sockets natively
-#![feature(unix_socket_abstract)]
```
BradLugo, line 34: I tried to understand the need for this, sifted through some commit history, but still don't quite grok it. This is the only code preventing cargo from building without the nightly flag.

BradLugo:
I thought I had signed the CLA. Is there a way to double check?

PR #17, opened 2022-09-19
Extract x509 cert data

future-highway:
Definitely needs more done, but it builds and the tests pass so far (I didn't figure out the signature encoding). 

There are more questions/comments that I had in the source file.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,7 +49,14 @@ serde = { version = "1.0.144", features = ["derive"] }
 tower = "0.4.13"
 home = "0.5.3"
 anyhow = "1.0"
-x509-certificate = "0.14.0"
+x509-parser = "0.14.0"
+rustls-pemfile = "1.0.1"
+if_chain = "1.0.2"
+time = "0.3.14"
+oid-registry = "0.6.0"
 
 [build-dependencies]
 tonic-build = "0.8"
+
+[dev-dependencies]
+openssl = "0.10.41"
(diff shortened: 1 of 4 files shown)


test_pki/_signed.client.nova.crt
```diff
@@ -0,0 +1,28 @@
+-----BEGIN CERTIFICATE-----
```
kris-nova, line 1: Having cert material in this repository makes me very nervous. It will only be a matter of time before this is picked up by a bad actor. Either someone malicious or someone running it production. 

Maybe we move the `test_pki` directory to something a little more inconspicuous? 

```
/assets/test/material/*.crt
/test/data/*.crt
/test/.data/*.crt
```

Something like that will make it a little less invasive 🙁

kris-nova (changes requested):
Can we please move the test data?

Otherwise the code looks great. Thanks for this contribution.

test_pki/_signed.client.nova.crt
future-highway: Yes, of course. These will be invalid certs though, right? Or I guess you are saying someone will mistakenly use them as valid certs?

future-highway:
So I ended up using the openssl crate and mimicked the makefile commands. Probably unnecessary as the cert data is never persisted, but I also made the validity period far in the future.

future-highway:
That probably pinged you Kris. Didn't mean to. I was trying to get the bot to change back to review needed. Guess that's not how. Sorry.

kris-nova:
No worries I have my GitHub notifications under control. Let me see the changes.

PR #19, opened 2022-09-22
Run clippy on build

future-highway:
Clippy has a default configuration, so it will output warnings for more than explicitly configured. This should be overridable, but starting with the defaults seems like a good choice.


kris-nova (approved):
Looks great. Thank you for your contribution!

PR #27, opened 2022-09-28
Added output macro

future-highway:
Converted the output code to a simple derive macro that will generate the output functions.


future-highway:
I'm renaming the macro lib, since it seems unwise to have named it after the only macro in it.

kris-nova (approved):

kris-nova:
I will fix it no worries!

PR #31, opened 2022-10-03
Getters/Setters Macro

future-highway:
Created a `Getters` and `Setters` macro to eliminate what looks like boilerplate code.

For now the macros only work on `struct` types with named fields (like the generated gRPC types). Fields can also have a `getset` attribute with options. Current options:
- `ignore_get` (don't generate getter)
- `ignore_set` (don't generate setter)
- `ignore` (don't generate getter or setter)

Usage looks like:
```rust
#[derive(macros::Getters, macros::Setters)
struct Executable {
    #[getset(ignore)]
    meta: Meta,
    name: String
}
```

There is also a [getset](https://docs.rs/getset/0.1.2/getset/) crate, but it didn't seem like it would match the project needs.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -62,11 +62,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
         )
         .type_attribute(
             "runtime.Executable",
-            "#[derive(serde::Serialize, serde::Deserialize)]",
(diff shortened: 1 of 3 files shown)


krisnova:
I love your work! Thank you SO much for doing this!

krisnova (approved):
Approve! We will need to add docs, but we can reference this PR for now.

PR #40, opened 2022-10-14
Improve Getters/Setters macros type compatibility

future-highway:
This should make the getters/setters support repeated fields with the exception of proto floats. I did light manual testing and it was working.

The pr can be merged if someone needs it (it is an improvement over the previous code), but I still need to organize/iterate it and test non-repeated types (e.g., non-repeated floats aren't currently panicking).

### DIFF
--- a/macros/src/get_set.rs
+++ b/macros/src/get_set.rs
@@ -1,0 +1,311 @@
+use proc_macro::TokenStream;
+use proc_macro2::Ident;
+use quote::{quote, ToTokens};
+use std::str::FromStr;
+use syn::parse::{Parse, ParseStream};
+use syn::punctuated::Punctuated;
+use syn::{parse_macro_input, Data, DeriveInput, Field, Token, Type};
+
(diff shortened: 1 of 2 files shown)


future-highway:
Updating based on 2nd commit:

The implementation should now be able to generate getters and setters, appropriate for use with rhai, for proto fields of the following proto types and their `repeated` variations:
- double
- int32, uint32, sint32, fixed32, sfixed32,
- int64, sint64, sfixed64
- bool
- string
- bytes
- message types should also work

The following proto types will panic on build:
- float (rhai only supports f64, and conversion to f32 will result in a loss of precision)
- uint64 (rhai integer types are i64, so u64 cannot be safely used without risking overflow)

Again, this pr can be merged if someone needs it (it is an improvement over the previous code), but I haven't looked into what happens when using `oneof` or and `enum` in the proto files.

krisnova (approved):

PR #51, opened 2022-10-17
stream logs via grpc api

Vincinator:
Adds a log target `StreamLogger` in addition to sending log messages to stdout/stderr.
This way, we do not depend on redirecting stdout/stderr of auraed for logging. (Thanks @MalteJ)
_No need for crate `gag`, or to wrap auraed in a Child Process_ 

log messages are send via a channel. A grpc API endpoint subscribes to a channel if a client has requested a stream. 
 
There are still some TODOs, major one is streaming stdout/stderr from subprocesses created via the exec function.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -155,6 +155,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "clap",
+ "crossbeam",
  "futures",
  "h2",
  "ipnetwork",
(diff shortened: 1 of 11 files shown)


krisnova:
There will be a lot of changes that I will need to make to this with regard to naming semantics and the type for a PID

However I think its best just to merge this in and iterate.

krisnova (approved):

Issue #53, opened 2022-10-18
GitHub actions seems to remove CNAME from pages settings and 404s


krisnova:
Upon merging a PR to main (or committing directly to the branch because I am a horrible person) a GitHub action is kicked off to update the static site using GitHub pages. Shortly after the event, the website [aurae.io](https://aurae.io) will return a 404. 

Looking in the get GitHub pages settings for the repository the domain name value is unexpectedly missing.

![image](https://user-images.githubusercontent.com/13757818/196443798-c0756e83-8b40-4e87-9fb8-257717037e2a.png)

Setting the value back to `aurae.io` fixes the 404 and the site is now updated with the most recent changes from the `main` branch.

krisnova:
Confirmed. Any merge or push to the main branch will break the website and result in a 404

![image](https://user-images.githubusercontent.com/13757818/196456110-f9aa5c72-91b2-4f14-93f2-10e518825fe1.png)

tcbyrd:
👋 from the Pages team. I was checking out your project and noticed this, so thought I'd offer a suggestion. It's now possible to use [Actions as the source](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-with-a-custom-github-actions-workflow), which allows you to have more control over the build. This option also removes the requirement for a CNAME file to exist, and you can deploy from any branch to Pages.

Happy to open a PR with an starter workflow to set this up if you like.

krisnova:
Also relevant to note that changing the `CNAME` file location to the configured directory Pages was looking in seemed to address this issue. I plan on making some changes to the GitHub documentation to clarify this as well.

https://github.com/aurae-runtime/aurae/pull/56 👀

krisnova:
Another related issue about helping myself and the maintainers keep an eye on things if the build breaks and the website goes down. https://github.com/aurae-runtime/aurae/issues/82

krisnova:
So the website has been down all morning. I checked the pages settings and it looks like the page settings had been reversed? All of our GitHub pages configuration was lost. 

https://github.com/aurae-runtime/aurae/settings/pages

I had to to back to the settings and re-enable the pages integration, select a branch, and re-enter the custom domain name. Waiting to see if the website comes back online.

krisnova:
Small update. In order to bring the site back online in its "final form" we need to point github pages at the `gh-pages` branch. This is the working config. Its been one of those exciting days!

[...]

PR #56, opened 2022-10-18
Add cname file for github pages

flostadler:
Added a CNAME file for github pages so deploys don't remove the custom domain name. This should fix the issue mentioned here https://github.com/aurae-runtime/aurae/issues/53

This explains why it's happening without the CNAME file: https://github.com/mkdocs/mkdocs/issues/1257

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-aurae.io
--- a/docs/CNAME
+++ b/docs/CNAME
@@ -1,0 +1,1 @@
+aurae.io


krisnova:
Would you mind also removing the `/CNAME` file in this PR as well?

flostadler:
> Would you mind also removing the `/CNAME` file in this PR as well?

Ah totally missed that the file already existed in the root of the project. Removed it now

krisnova:
Welcome to Aurae! We hope you like it here.

Issue #54, opened 2022-10-18
Break the build on compile warnings


krisnova:
Can we please break the build and fail a pull request if code creates a rust warning? 

We are trying to keep a clean set of code for the project, and we would like to leverage the Makefile commands to ensure that no warnings are generated. 

This includes 

 - Documentation
 - Linting
 - Unused (dead) code
 - etc

dominichamon:
i'll start on this to get a feel for the code if you want

future-highway:
Thinking about the approach to this, I'm not sure about using `deny` to fail the build. It makes it more difficult to differentiate normal errors from lint based ones. I'm also concerned about discouraging contributions since not everyone is fully comfortable with rust yet.

An alternate approach is to use `warn` for any lints we want (that are default-allow). Then, we can alter the `cargo clippy` to `cargo clippy -- -D warnings` in the GitHub checks. The hope is that we don't discourage contributors from opening a PR, where those who need it can get a review/help in fixing the warnings.

dominichamon:
ok, that's fair. i think we'd also want to add this to the `Makefile`, not just the github checks.

i can start by adding the warnings to the Makefile and fixing anything, then we can audit any default-allow warnings we want to add?

future-highway:
I'm fine with whatever the decision is, but the reason I didn't want to suggest adding it to the makefile was to let people open PR's with the warnings. `cargo clippy -- -D warnings` will make all the warnings show as errors. 

Maybe we can have an additional makefile command that is strict and people could be encouraged to run it prior to the PR, but wouldn't be part of the normal dev flow.

bpmooch:
at work we have below and it works pretty good

```
install: deps
	@cargo install --path .

## Lint code
lint: deps
	@rustup component add clippy 2> /dev/null
	@cargo clippy --all-features --all -- -D clippy::all -D warnings
```

dmah42:
i think it's ok to close this. we have some more to enable but it doesn't need an issue to track :)

PR #57, opened 2022-10-19
Warnings as errors

dominichamon:
I wanted to enable them all, but it will require some more work (so i won't mark #54 as fixed).

i also thought about creating an attribute that can be shared between the various crate roots but maybe that's overkill.  this implementation allows someone to fix one warning at a time.

### DIFF
--- a/auraed/src/bin/main.rs
+++ b/auraed/src/bin/main.rs
@@ -29,6 +29,34 @@
 \* -------------------------------------------------------------------------- */
 
 #![warn(clippy::unwrap_used)]
+#![deny(bad_style,
+        const_err,
+        dead_code,
+        improper_ctypes,
+        non_shorthand_field_patterns,
+        no_mangle_generic_items,
+        overflowing_literals,
+        path_statements,
+        patterns_in_fns_without_body,
+        private_in_public,
+        unconditional_recursion,
+        // TODO: unused,
+        // TODO: unused_allocation,
+        // TODO: unused_comparisons,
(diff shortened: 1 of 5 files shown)


auraed/src/bin/main.rs
```diff
@@ -29,6 +29,34 @@
 \* -------------------------------------------------------------------------- */
 
 #![warn(clippy::unwrap_used)]
+#![deny(bad_style,
+        const_err,
```
future-highway, line 33: This is [deny-by-default](https://doc.rust-lang.org/rustc/lints/listing/deny-by-default.html), so doesn't need to be included

auraed/src/bin/main.rs
```diff
@@ -29,6 +29,34 @@
...
+        dead_code,
+        improper_ctypes,
+        non_shorthand_field_patterns,
+        no_mangle_generic_items,
+        overflowing_literals,
+        path_statements,
+        patterns_in_fns_without_body,
```
future-highway, line 40: This is [deny-by-default](https://doc.rust-lang.org/rustc/lints/listing/deny-by-default.html), so doesn't need to be included

future-highway:
A couple of the lints were deny-by-default, so commented on those, but not on all occurrences (figured being repetitive  wasn't needed).

krisnova (approved):

krisnova:
Welcome to Aurae. We hope you like it here.

PR #61, opened 2022-10-20
Makefile lint

dominichamon:
A less aggressive approach to #54 that enables a `make lint` command that is also run in the github workflows, but remains optional for local development.

Also enables (and fixes) a few non-default warnings.


krisnova:
I love the idea of pulling these apart so lint is its own make target now. I also think it might be helpful to come up with yet another way of letting the developer know that a lint will fail their build/PR if it's not ran. Maybe a git hook? But that also could get noisy. Either way for now this is a great step.

krisnova (approved):

PR #76, opened 2022-10-23
Introduce cgroup management for executables

krisnova:
This pull request adds a cgroup layer for all executables and begins to establish the `meta.name` field as the unique identifier for workloads in Aurae.


auraed/Cargo.toml
```diff
@@ -64,9 +64,10 @@ rtnetlink = "0.11.0"
 netlink-packet-route = "0.13.0" # Used for netlink_packet_route::rtnl::address::nlas definition
 thiserror = "1.0.37"
 crossbeam = "0.8"
-libcontainer     = { git = "https://github.com/containers/youki.git", tag = "v0.0.2" }
-liboci-cli     = { git = "https://github.com/containers/youki.git", tag = "v0.0.2" }
-
+libcontainer = { git = "https://github.com/containers/youki.git", tag = "v0.0.2" }
```
dominichamon, line 67: why do we depend on this same repo twice with different names?

future-highway: When you rely on a git repo, cargo will look for a crate with the name you use, in that repo (not the whole repo). So here, there are 2 different crates being used that happen to be in the same repo. [Docs](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)

dominichamon: ah thanks :) #til

krisnova:
Notes to address: 

Some question/thoughts re the current api:

1) Are we going to be keeping the current request/response types? Or wrapping them in, for example RunExecutableRequest/RunExecutableResponse? It makes more sense to me to have AuraeMeta (and other metas) on the wrapping types.
2) I'm not sure how I feel about the meta "pattern". Does it run the risk of becoming a junk drawer for fields that get tacked on (and consequently appearing on every req/res)?
3) I get that AuraeMeta.name maps to the cgroup name. There was also mention of "buckets"/"namespace" for multitenancy. Where does that come? mTLS certs?

auraed/src/lib.rs
```diff
@@ -225,6 +227,44 @@ impl AuraedRuntime {
...
+    command: String,
+    timestamp: SystemTime,
+}
+
+/// A deterministic function used to calculate a meaningful cgroup directory
+/// name from an arbitrary executable shell string.
+/// Note: The return value of this function is how we look up Aurae cells
+/// Note: from other subsystems
+fn cell_name_from_string(command: &str) -> Result<String, anyhow::Error> {
+    let now = SystemTime::now();
```
dominichamon, line 242: this is not idempotent so the cell name will not be deterministic (per the comment).  or did i misunderstand _how_ deterministic you want this to be?

auraed/src/runtime/mod.rs
```diff
@@ -56,58 +57,88 @@ pub struct CoreService {}
...
+        let cell_id = cell_name_from_string(&r.command)
+            .expect("cell name hash calculation");
+
+        info!("Spawning cell: {}", cell_id);
+
+        // Create the cgroup for the cell
+        let hierarchy = cgroups_rs::hierarchies::auto(); // v1/v2 cgroup switch automatically
+        let cgroup: Cgroup = CgroupBuilder::new(&cell_id)
+            .cpu()
+            .shares(10) // Use 10% of the CPU relative to other cgroups
```
dominichamon, line 82: should this be part of the Executable Request?

i imagine memory could be too (and maybe networking?)

dominichamon: having it part of the Request would also make it easier to build a scheduling layer on top, if the "available CPU" is exposed for a Cell

auraed/src/runtime/mod.rs
```diff
@@ -56,58 +57,88 @@ pub struct CoreService {}
...
+                let proc = meta::ProcessMeta { pid: pid as i32 };
+                let status = meta::Status::Complete as i32;
+
+                // Parse stdout from pipe
+                let stdout_val = String::from_utf8_lossy(&output.stdout);
+
+                // Parse stderr from pipe
+                let stderr_val = String::from_utf8_lossy(&output.stderr);
+
+                let response = ExecutableStatus {
```
dominichamon, line 128: we only respond after the command has completed?  it would be nicer to respond before completion, perhaps with a streaming response for logging.

PR #85, opened 2022-10-24
example scripts were not compatible with the latest changes of runtim…

adililhan:
due to changes in runtime (auraescript), scripts throw some exceptions. exceptions fixed by this PR 🎉


krisnova (approved):
Welcome to Aurae! We hope you love it here!

PR #89, opened 2022-10-25
enable unused comparisons warning

dominichamon:


krisnova (approved):
Keep them coming let's gooo!

PR #90, opened 2022-10-25
Fixpod cell

dominichamon:


MalteJ (approved):

krisnova:
Did we make the decision to have a single cell per pod? Just wanted to make sure we thought this through. I have weak opinions here (for now).

MalteJ:
I am in favor of exactly one Cell per Pod. Otherwise you create a new "node" which has a certain amount of ressources and you have to schedule cells into this "node". You'd get constraints on scheduling, which would in the end result in lower utilization of your overall compute ressources.
The overhead of one MicroVM per cell should be small in comparison. Especially when using a specialized Kernel and an aurae initramfs.
Also one cell per pod is less complex from a security perspective. You'd simply work with pods and know they are isolated using virtualization. If you have a super special use case, where you have to put multiple cells into one MicroVM, you could still achieve this by calling `Spawn()` and manage the spawned aurae MicroVM on your own.

dominichamon:
this came out of a discussion on the discord in which i was persuaded that this was what was intended.

i have no strong opinion.

MalteJ:
by the way, with grpc you could expand the API later on easily to support multiple cells per pod.

krisnova:
I am convinced that a single cell per pod is the future. After some conversations at KubeCon I think I might add some additional functionality to cells in the #26 PR. This will reinforce this PR. 

Good call out @dominichamon and @MalteJ

krisnova (approved):

PR #92, opened 2022-10-26
fix make clean

fdr10:
## What it does ?
fixes error when running the clean target in the top level Makefile
## How it does it ?
In the top level Makefile replace `cd aurae && make clean` with `cd auraescript && make clean`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,11 +113,15 @@ clean-certs: ## Clean the cert material
 fmt: headers ## Format the entire code base(s)
 	@./hack/code-format
 
+clean-auraescript:
+	cd auraescript && make clean 
+
+clean-auraed:
+	cd auraed && make clean
+
 .PHONY: clean
 clean: clean-certs
-	cd auraed && make clean
-	cd auraescript && make clean
-	@rm -rvf target/*
+	cargo clean
 
 headers: headers-write ## Fix headers. Run this if you want to clobber things.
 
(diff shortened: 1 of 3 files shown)


fdr10:
@cla-bot check

fdr10:
Hello ! This is my first open source pr. Looking forward to contribute more as I get familiar with the project. Starting small. 🙂

Makefile
```diff
@@ -115,8 +115,8 @@ fmt: headers ## Format the entire code base(s)
 
 .PHONY: clean
 clean: clean-certs
-	cd aurae && make clean
 	cd auraed && make clean
+	cd auraescript && make clean
 	@rm -rvf target/*
```
dominichamon, line 120: this is ok, but i wonder if we should _also_ call `cargo clean` as we do in the sub-folder Makefiles.

krisnova: What does the nested makefile do? Doesn't that itself call cargo clean?

fdr10: Hi everyone 🙂. Yeah they call `cargo clean`.  This got me thinking if instead we should replace the clean target with a `cargo clean` (which also does an `rm -rf /target`)   while at the same time leverage the existing Makefiles to clean the packages (auraed and auraescript) individually if needed.

krisnova: Sure however you want to do it. I will merge now.

krisnova (approved):

PR #94, opened 2022-10-27
chore(refactor): move protos into api library

edude03:
Todos: 
- [ ] Lift macros up to the top level
- [ ] fix references in auraescript
- [ ] fix the build script for api


auraed/src/init/system_runtime.rs
```diff
@@ -3,7 +3,7 @@ use crate::init::network::{
 };
 use crate::init::power::spawn_thread_power_button_listener;
 use crate::init::{fs, logging, InitError, BANNER};
-use crate::observe::LogItem;
+use api::observe::LogItem;
```
edude03, line 6: Can change this back

PR #95, opened 2022-10-27
feat(nix): Add support for building/running with nix

edude03:


krisnova (approved):
I don't know enough about Nix - but I do know enough about @edude03 to know this is trustworthy. LGTM.

flake.nix
```diff
@@ -0,0 +1,80 @@
...
+    flake-utils.follows = "cargo2nix/flake-utils";
+    nixpkgs.follows = "cargo2nix/nixpkgs";
+  };
+
+  outputs = inputs: with inputs; # pass through all inputs and bring them into scope
+
+    # Build the output set for each default system and map system sets into
+    # attributes, resulting in paths such as:
+    # nix build .#packages.x86_64-linux.<name>
+    flake-utils.lib.eachDefaultSystem (system:
```
hazelweakly, line 21:
```suggestion
    flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ] (system:
```

Aurae can't be compiled on non-linux (yet?) and setting up cross compiling was an absolutely massive pain that I didn't manage to get working, so we can shelve that for now. But, this probably shouldn't claim to build on darwin in that case :)

PR #96, opened 2022-10-27
More efficient CI

future-highway:
Attempting to speed up the GitHub checks that are run on pr's.

- Don't `cargo install` auraed or auraescript, as we don't need to do that for the checks
- Run build before lint since `cargo clippy` can leverage the output of `cargo build` seemingly better than vice versa.

Also added a `make prcheck` that runs the same checks that gh runs, for convenience.


bpmooch (approved):
looks litaf to me (lltm)

dominichamon (approved):

PR #97, opened 2022-10-28
More refactoring of init module (mainly network)

future-highway:
Refactored the network code to better contain it in its own module. 

Also eliminated some errors from `init/fs`, as we currently control the input and can use `expect` instead.

Note: Network can probably be improved if the link indexes are stable across calls, but I didn't know if that is true, so did not make that assumption.

### DIFF
--- a/auraed/src/init/fs.rs
+++ b/auraed/src/init/fs.rs
@@ -1,55 +1,30 @@
 use log::{error, info};
-use std::ffi::{CString, NulError};
+use std::ffi::{CStr, CString};
 use std::ptr;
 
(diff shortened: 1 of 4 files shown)


krisnova:
LGTM but leaving for @Vincinator to merge/review

Vincinator (approved):
Thanks - looks good!

future-highway:
@Vincinator thanks for the review. Resolving the conflicts removed your review for the merge. If you are still around to give it a quick approve, great. No rush though.

Vincinator (approved):
Hey, @future-highway - this is an easy approve. Thank you!

PR #98, opened 2022-10-28
Replace crossbeam::mpmc with tokio's broadcast

future-highway:
Tokio's broadcast supports async/await

### DIFF
--- a/auraed/src/logging/logchannel.rs
+++ b/auraed/src/logging/logchannel.rs
@@ -32,8 +32,8 @@
 #![allow(dead_code)]
 //
 use crate::observe::LogItem;
-use crossbeam::channel::{bounded, Receiver, Sender};
 use log::error;
+use tokio::sync::broadcast::{self, Receiver, Sender};
 
 use super::get_timestamp_sec;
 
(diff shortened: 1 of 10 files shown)


auraed/src/logging/logchannel.rs
```diff
@@ -60,28 +59,24 @@ impl LogChannel {
...
-        match producer.send(LogItem {
+        // send returns an Err if there are no receivers. We ignore that.
+        let _ = producer.send(LogItem {
             channel: "unknown".to_string(),
             line: line.to_string(),
             // TODO: milliseconds type in protobuf requires 128bit type
             timestamp: get_timestamp_sec(),
-        }) {
-            Ok(_) => {}
-            Err(e) => {
```
dominichamon, line 75: are we losing some error logging with this change?

future-highway: Yes and no. LogChannel will only send messages while there is a receiver (connected gRPC client). Clients only get messages that are produced after they connect/subscribe. So if there is no receiver, LogChannel will lose messages.  However, SimpleLogger is also wired up, and that will receive all log messages I believe.

krisnova (approved):
Works for me

future-highway:
I'm resisting the urge to press the merge button, to let Vincinator get a chance to review/comment.

But if anyone else wants to press the button, I won't object, lol.

krisnova (approved):

PR #99, opened 2022-10-28
executables modified to handle the stderr output produced by wget

adililhan:
I have tried to run a wget script in Aurae. However, it never ends. I think it is because `wget` keeps producing output in `stderr`

I modified the code. I've also added a new test file to make sure the code can handle `wget`

I'd like to hear the feedbacks.

Thanks

### DIFF
--- a/auraed/src/runtime/mod.rs
+++ b/auraed/src/runtime/mod.rs
@@ -41,7 +41,6 @@ use cgroups_rs::cgroup_builder::CgroupBuilder;
 use cgroups_rs::*;
 use cgroups_rs::{CgroupPid, Controller};
 use log::{debug, info};
-use std::io::Read;
 use std::process::Stdio;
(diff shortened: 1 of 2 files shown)


future-highway:
Looking at these changes and the source for `Command`, I think the main change is that the `wait` in `wait_with_output` occurs after reading stdout/stderr.

`String::from_utf8_lossy` is also a good idea, as we wouldn't want to panic just because the output is not UTF8, I don't think.

Since there is an open PR being worked on by @krisnova with a bunch of changes to this file, I'll skip marking it as reviewed so she can take a look and decide what to do.

krisnova (approved):

krisnova:
I went ahead and merged it. I will rebase #26 on top of this, and ensure the wget script works in my new feature. Now big deal.

PR #103, opened 2022-10-30
enable the trivial casts lint warning

dominichamon:
this one required some work to avoid a cast and use coercion instead

i also fixed a use of unwrap. i know it's ok to use in tests, but clippy was yelling at me.

### DIFF
--- a/auraed/src/init/power.rs
+++ b/auraed/src/init/power.rs
@@ -63,6 +63,10 @@ pub(crate) struct InputEvent {
 // see  https://elixir.bootlin.com/linux/latest/source/include/uapi/linux/input-event-codes.h#L191
 const KEY_POWER: u16 = 116;
 
+fn to_u8_ptr<T>(p: *mut T) -> *mut u8 {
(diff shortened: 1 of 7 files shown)


auraed/src/init/power.rs
```diff
@@ -88,10 +92,7 @@ pub(crate) fn spawn_thread_power_button_listener(
     std::thread::spawn(move || {
         loop {
             let event_slice = unsafe {
-                slice::from_raw_parts_mut(
-                    &mut event as *mut _ as *mut u8,
```
future-highway, line 92: I wonder if you could have removed one of the mut's here and have had it work?

future-highway (approved):

auraed/src/init/power.rs
dominichamon: i tried a few different things with the cast, but only the coercion seemed to work.  there is an inline version with two `let` statements but it's uglier than the method, i think.

PR #114, opened 2022-11-10
Capture os error as rust error

future-highway:
According to the [mount docs](https://man7.org/linux/man-pages/man2/mount.2.html#RETURN_VALUE), on error it will return -1 and set the errno to indicate the error.

[io::Error::last_os_error](https://doc.rust-lang.org/std/io/struct.Error.html#method.last_os_error) "reads the value of errno for the target platform (e.g. GetLastError on Windows) and will return a corresponding instance of [Error](https://doc.rust-lang.org/std/io/struct.Error.html) for the error code."

### DIFF
--- a/auraed/src/init/fs.rs
+++ b/auraed/src/init/fs.rs
@@ -1,11 +1,16 @@
 use log::{error, info};
-use std::ffi::{CStr, CString};
-use std::ptr;
+use std::ffi::CStr;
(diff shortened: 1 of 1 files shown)


dominichamon (approved):

dominichamon:
thanks!

krisnova (commented):
Loots good

PR #115, opened 2022-11-16
Introduce Deno instead of Rhai (Make AuraeScript also support TypeScript)

krisnova:
This will introduce Deno for the main scripting core such that AuraeScript is now valid JavaScript/TypeScript. 

We can plumb the gRPC magic through the remaining `/legacy` directory as an example. 

 - [ ] Plumb mTLS back through config
 - [ ] Add command line parlance for scripting interpreter
 - [ ] Add object/cell examples
 - [ ] Figure out documentation


krisnova:
Closed in favor of #120

PR #116, opened 2022-11-17
Validation

future-highway:
WARNING: This PR introduces validation and validation_macros crates which are opinionated in their expectations of the conventions we should use but makes scaffolding code easier ONLY IF you have a cooperative IDE.

Some expectations: 
- incoming messages should be validated ASAP.
- we should fail fast (return on first validation error)
- validating the incoming message should return a new type
- the new type will have the same fields, but the field types can, and often should, be different

---

As part of the PR, I have scaffolded validating the `allocate` endpoint on the `CellService`. Steps I had to take.

1. Create a validated struct based on the unvalidated proto message (`field_type` is needed because the type differs from the unvalidated type):
```rust
#[derive(ValidatingType)]
pub(crate) struct ValidatedAllocateCellRequest {
    #[field_type(Option<Cell>)]
    cell: ValidatedCell,
}
```

2. The macro generated the `AllocateCellRequestTypeValidator` trait, and an empty `AllocateCellRequestValidator` struct so I wrote:
```rust
impl AllocateCellRequestTypeValidator for AllocateCellRequestValidator {
    
}
```

3. CLion being a cooperative IDE (hopefully VSC will also be), it provided a "quick-fix" action to generate:
```rust
impl AllocateCellRequestTypeValidator for AllocateCellRequestValidator {
    fn validate_cell(
        cell: Option<Cell>,
        field_name: &str,
        parent_name: Option<&str>,
    ) -> Result<ValidatedCell, ValidationError> {
        todo!()
    }
}
```

4. Implement the todo
```rust
impl AllocateCellRequestTypeValidator for AllocateCellRequestValidator {
    fn validate_cell(
        cell: Option<Cell>,
        field_name: &str,
        parent_name: Option<&str>,
    ) -> Result<ValidatedCell, ValidationError> {
        let cell = validation::required(cell, field_name, parent_name)?;

        ValidatedCell::validate(
            cell,
            Some(&validation::field_name(field_name, parent_name)),
        )
    }
}
```

5. We need a `ValidatedCell` so (leaving out fields to be succinct here):
```rust
#[derive(ValidatedType)]
struct ValidatedCell {
    name: String,
    cpus: String,
}
```

6. Repeat step 2 but with `CellTypeValidator` and `CellValidator`, then step 3 and 4.

---

 Things to note:
- There are 2 derive macros (`ValidatingType`, `ValidatedType`). The only difference is which type has the `validate` function. 
- You can skip decorating fields with the `field_type` attribute if the validated type is exactly the same as the unvalidated type, but Rust has a great type system, so making types such as `struct CellName(String)` can help reduce errors in code and provides a place to write reusable code. Ex:
```rust
struct CellName(String);

impl validation::ValidatedField<String> for CellName {
    fn validate(
        input: Option<String>,
        field_name: &str,
        parent_name: Option<&str>,
    ) -> Result<Self, ValidationError> {
        let input = validation::required_not_empty(
            input,
            field_name,
            parent_name,
        )?;

        validation::allow_regex(
            &name,
            &validation::DOMAIN_NAME_LABEL_REGEX,
            field_name,
            parent_name,
        )?;

        Ok(CellName(input))
    }
}
```

We could then adjust ValidatedCell to:
```rust
#[derive(ValidatedType)]
struct ValidatedCell {
    #[field_type(String)]
    name: CellName,
    cpus: String,
}
```

and implement validation like:

```rust
impl CellTypeValidator for CellValidator {
    fn validate_name(
        name: String,
        field_name: &str,
        parent_name: Option<&str>,
    ) -> Result<CellName, ValidationError> {
        CellName::validate(Some(name), field_name, parent_name)
    }

    fn validate_cpus(
        _cpus: String,
        _field_name: &str,
        _parent_name: Option<&str>,
    ) -> Result<String, ValidationError> {
        todo!()
    }
}
```


future-highway:
Another note...

You can see in the "Adjust validation macros" commit, I made changes that allow multiple validated structs to exist in the same module, which I took advantage of for `ValidatedAllocateCellRequest` and `ValidatedCell`. While this is convenient (read "less annoying when developing"), I would say it should generally be avoided. `ValidatedCell` should be put in its own module, but we can easily do that in refactoring PRs.

dominichamon:
it's not entirely clear to me how much this is going to slow down development or get ignored during initial development, but i really like the goal and the implementation is pretty clean.

future-highway:
I suspect it runs a high risk of both of those, which is part of why I want to be carful in merging this.

I do think we need a solution, but this may not be the right one, or it may not be the right time for it.

I'm also curious if VS Code (as I assume that will be the most common editor) triggers quick-fixes in this scenario. If not, this approach will be VERY annoying, and would guarantee slowdown/ignoring it.

future-highway:
I expanded the macro's capabilities to also potentially generate the trait's validation function:

```rust
#[derive(ValidatingType)]
pub(crate) struct ValidatedExampleRequest {
    #[validate]
    #[field_type(String)]
    auto_validate_by_calling_validate_on_my_type: MyType,

    #[validate(opt)]
    #[field_type(Option<String>)]
    auto_validate_by_calling_validate_optional_on_my_type: MyType,

    #[validate(none)]
    don't_validate_just_pass_through: String
}
```

The first variation (`#[validate]`) assumes the input IS NOT optional so generates code that looks like: `MyType::validate(Some(input), field_name, parent_name)`.

The second variation (`#[validate(opt)]`) assumes the input IS optional so generates code that looks like: `MyType::validate_optional(input, field_name, parent_name)`.

The third variation (`#[validate(none)]`) skips validation, so the input and output types must match, generating: `Ok(input)`.

krisnova (approved):

PR #118, opened 2022-11-19
enable unused_results

dmah42:
enable this everywhere, with explicit `_unused` variables where necessary.  this is good as it helps document intent.

auraescript is not included here because rhai engine creation is going to be ripped out and that's the main blocker to enabling the warning.

### DIFF
--- a/auraed/src/init/network/mod.rs
+++ b/auraed/src/init/network/mod.rs
@@ -73,7 +73,7 @@ pub(crate) struct Network(Handle);
 impl Network {
     pub(crate) fn connect() -> Result<Network, NetworkError> {
         let (connection, handle, _) = rtnetlink::new_connection()?;
-        let _unused_join_handle = tokio::spawn(connection);
+        let _ = tokio::spawn(connection);
         Ok(Self(handle))
     }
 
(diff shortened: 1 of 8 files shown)


auraed/src/init/network/mod.rs
```diff
@@ -73,7 +73,7 @@ pub(crate) struct Network(Handle);
 impl Network {
     pub(crate) fn connect() -> Result<Network, NetworkError> {
         let (connection, handle, _) = rtnetlink::new_connection()?;
-        tokio::spawn(connection);
+        let _unused_join_handle = tokio::spawn(connection);
```
future-highway, line 76: using `let _ =` is fine. You don't need to spend energy naming things that are ignored.

auraed/src/init/power.rs
```diff
@@ -37,7 +37,9 @@ use ::libc;
 #[allow(dead_code)]
 pub(crate) fn syscall_reboot(action: i32) {
     unsafe {
-        libc::reboot(action);
+        if libc::reboot(action) != 0 {
+            panic!("failed to reboot");
```
future-highway, line 41: I don't think we want to panic. However, power.rs needs other work anyway. If you make other changes to this PR before merging, throw in a "TODO: Handle this better", just to make sure it doesn't get overlooked.

future-highway (dismissed):
No real issues, so I'm going to approve and leave it up to you if you want to make the changes from the comments or just merge.

future-highway (approved):

PR #120, opened 2022-11-20
Auraescript on Deno

future-highway:
Current state:
- op_decl are generated with code similar to what we had with rhai (see runtime/mod.rs)
- build.rs will generate TypeScript files based on our proto files  (e.g., lib/runtime.ts) using ts-proto library from npm
- macros generate op_decls and augment the generated typescript files with clients that call the appropriate deno ops.
- TS script can import the generated typescript files (see auraescript_deno/examples/script.ts)
- confirmed can execute TS scripts (with script.ts a call is successfully sent to auraed which errors with not implemented for `CellService.allocate`)
- did not confirm JS script works

I named the crates "auraescript_deno" and "auraescript_deno_macros" just so that they don't interfere what we have or other PR's. However auraescript_deno's bin is still named auraescript


dmah42:
how does this overlap with https://github.com/aurae-runtime/aurae/tree/deno ?

dmah42:
> how does this overlap with https://github.com/aurae-runtime/aurae/tree/deno ?

discussed in the discord: this is the way.

auraescript/Cargo.toml
```diff
@@ -1,69 +1,25 @@
...
-#   See the License for the specific language governing permissions and        #
-#   limitations under the License.                                             #                                                                             #
-#                                                                              #
-# ---------------------------------------------------------------------------- #
-
 [package]
 name = "auraescript"
 version = "0.1.0"
 edition = "2021"
-authors = ["The Aurae Authors", "Kris Nóva <kris@nivenly.com>"]
```
dmah42, line 35: might be good to keep this and the above license notice

auraescript/LICENSE
```diff
@@ -1,201 +0,0 @@
-                                 Apache License
```
dmah42, line 1: why did this get deleted?

auraescript/Makefile
```diff
@@ -41,7 +41,7 @@ compile: ## Compile for the local architecture ⚙
 install: ## Build and install (debug) 🎉
 	@$(cargo) clippy
 	@$(cargo) install --debug --path .
-	sudo -E cp -rv examples/bin/* /usr/bin
+	#sudo -E cp -rv examples/bin/* /usr/bin
```
dmah42, line 44: i guess either put this back or remove the line?

auraescript/README.md
```diff
@@ -1,62 +0,0 @@
-# AuraeScript
```
dmah42, line 1: is this not relevant any more?

auraescript/src/bin/main.rs
```diff
@@ -1,169 +1,32 @@
...
- *   Unless required by applicable law or agreed to in writing, software      *
- *   distributed under the License is distributed on an "AS IS" BASIS,        *
- *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
- *   See the License for the specific language governing permissions and      *
- *   limitations under the License.                                           *
- *                                                                            *
-\* -------------------------------------------------------------------------- */
-
-// TODO @kris-nova as we move to Deno we probably want to revist the main function
-// Lint groups: https://doc.rust-lang.org/rustc/lints/groups.html
```
dmah42, line 32: do you mind putting these back? they shouldn't be hard to resolve i hope...

auraescript/LICENSE
future-highway: By accident, but I'm not sure if we still need it. Is this an artifact of when we had multiple repos. Put back for now.

auraescript/README.md
future-highway: I put it back and made some updates to make it more inline with what we have.

auraescript/Cargo.toml
```diff
@@ -1,69 +1,26 @@
...
+thors = ["The Aurae Authors", "Kris Nóva <kris@nivenly.com>"]
```
dmah42, line 5: "we're still worthy!" but i think this should be "authors"

auraescript/src/bin/main.rs
future-highway: Put back, by I had to suppress a little as there is dead code that we will probably use and I couldn't get tonic to take the attributes on the generated code

auraescript/Cargo.toml
future-highway: of course I messed up copy pasting, lol

dmah42 (approved):

dmah42 (approved):

PR #121, opened 2022-11-21
Auraescript fix

future-highway:
There were a couple of bugs that this should fix
- All generated TS Client classes were implementing `CellService` (made that dynamic)
- The macro would often append the generated TS client multiple times as cargo will rerun only parts of the build/code generation. Changed the implementation to have build.rs output to `lib/temp` directory and then have the macro output to `lib` where it overrides the entire file instead of appending.


future-highway:
I don't know what the magic combinations of strings are to keep the `lib` and `lib\temp` directories without committing their contents (which is why `lib\temp\runtime.ts` is slipping in).

auraescript/lib/temp/runtime.ts
```diff
@@ -0,0 +1,431 @@
+/* eslint-disable */
```
krisnova, line 1: Did you want `temp` to be in the PR?

future-highway: There is definitely a cleaner organization possible so that the temp dir won't be needed either, but I wanted to get the fix in ASAP

future-highway: Looks like I figured it out. Will try to eliminate the intermediary files (temp) being output in a future refactor. Probably after the holiday.

krisnova (approved):
Beautiful code as per usual.

PR #133, opened 2022-11-28
Auraescript refactoring

future-highway:
Besides other changes, this PR includes regenerating proto files on `build` (introduces a dependency on buf in CI) and other appropriate makefile commands. I didn't think depending on contributors to always remember to rebuild the latest protos prior to creating a PR was a good idea.

### DIFF
--- a/.github/workflows/build-compile-test.yml
+++ b/.github/workflows/build-compile-test.yml
@@ -11,6 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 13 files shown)


future-highway:
I'm now doubting my decision to have build trigger buf. As we don't have any tests that cover that code yet, I don't think it makes a difference in catching PR mistakes.

Another option is to not have it trigger anywhere CI hits (i.e., remove `proto` call from `build` and put it in `prcheck`).

I'll leave the decision up to whoever reviews this.

krisnova (approved):
LGTM

PR #139, opened 2022-12-12
introduce real error handling

dmah42:


dmah42:
putting this up to get comments from @krisnova and @future-highway on whether this is the way we want to think about mapping errors to tonic::Status or if there's a better way.

future-highway:
Kind of...

First thing is that a ValidationError -> Status::failed_precondition will be written once (at least that's the goal) and we won't have to worry about it passed calling `.validate`.

Second thing is that I think we will end up removing all (or nearly all) the `.expect` calls.

I haven't formed enough of an opinion on the errors that arise from the core logic (though I'm pretty sure we will use the `thiserror` crate). I tend to like the core logic to not even being aware of the proxy (is that the right word, I'm referring to whether it be grpc or rest or whatever), which may necessitate creating an error enum per endpoint and the impl `From<EndpointError> to Status`.  This is again a choice that I think results in better (read, more maintainable) code, but has a bit more overhead for development.

You might be catching onto a pattern with my suggestions...I like to macro away the boilerplate, but spend extra time being as explicit as possible or leveraging the type system otherwise.

dmah42:
> Kind of...
> 
> First thing is that a ValidationError -> Status::failed_precondition will be written once (at least that's the goal) and we won't have to worry about it passed calling `.validate`.

that would be great and gets rid of two places where i'm checking that an optional request parameter is set.

> 
> Second thing is that I think we will end up removing all (or nearly all) the `.expect` calls.

that's the goal of this PR (should be now complete)

> 
> I haven't formed enough of an opinion on the errors that arise from the core logic (though I'm pretty sure we will use the `thiserror` crate). I tend to like the core logic to not even being aware of the proxy (is that the right word, I'm referring to whether it be grpc or rest or whatever), which may necessitate creating an error enum per endpoint and the impl `From<EndpointError> to Status`. This is again a choice that I think results in better (read, more maintainable) code, but has a bit more overhead for development.

i recently went through this in another project and settled on the anyhow crate over thiserror (and they don't play nice together, and aurae already uses anyhow elsewhere).  supposedly anyhow is good for applications and thiserror is good for libraries, but i don't know how grpc mixes that up.  actiserve, for example, is an application but thiserror works better as the server response status just fits that model more.

https://github.com/hachyserve/actiserve/blob/main/src/error.rs is where i think we'll end up and this PR is step 1 towards that. 

> 
> You might be catching onto a pattern with my suggestions...I like to macro away the boilerplate, but spend extra time being as explicit as possible or leveraging the type system otherwise.

dmah42:
i should note: if we're going to go with thiserror then i am happy to add that to this PR but we should think about getting rid of anyhow elsewhere i think.

future-highway:
Just double checked. I think anywhere you see `anyhow` used in auraed, just consider it in need of refactoring out/legacy. Some of the previous PRs targeting the `init` module were mainly organizing code and replacing `anyhow` with `thiserror`.

We don't need to cover it all in one PR, IMO. I know we want to keep the main branch at a certain level of quality, but I think we have the reality of just starting to have things form.

Also, I think when they said anyhow is good for applications, they were thinking more of CLI type apps, because I haven't found it to be true if you just think about bin vs lib (which was how I originally read it).

PR #142, opened 2022-12-12
Initial validation integration step

future-highway:
This only integrates validation for CellService.free as a first step to show the pattern.

Things that changed:
- We now have a dedicated type for cell_name, `CellName`, with centralized validation. Ideally we shouldn't be doing strict validation of `cell_name` on `free` (required_not_empty may be enough), and only run the stricter validation on `allocate` when creating the cgroup, but I wanted to show more of how things work.
- The `free_cell` endpoint has its own file where the related logic can all exist together. I think this leads to cleaner code.
- If you look at the code in CellService.free, you can see how it's essentially becoming boilerplate (besides the `info!`) and we can write a macro to write it for us as long as we keep a consistent signature on `handle`.

### DIFF
--- a/auraed/src/runtime/cell_name.rs
+++ b/auraed/src/runtime/cell_name.rs
@@ -1,0 +1,38 @@
+use std::ops::Deref;
+use validation::ValidationError;
+
+#[derive(Debug)]
+pub(crate) struct CellName(String);
+
+impl validation::ValidatedField<String> for CellName {
+    fn validate(
+        input: Option<String>,
+        field_name: &str,
+        parent_name: Option<&str>,
+    ) -> Result<Self, ValidationError> {
+        let input =
+            validation::required_not_empty(input, field_name, parent_name)?;
+
+        // TODO: what makes a valid cgroup name
+        // any valid path?
+        // do we want a restriction on length?
+        // anything else?
+        // Highly restrictive for now:
+        validation::allow_regex(
+            &input,
+            &validation::DOMAIN_NAME_LABEL_REGEX,
+            field_name,
+            parent_name,
+        )?;
+
+        Ok(Self(input))
+    }
+}
+
+impl Deref for CellName {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
--- a/auraed/src/runtime/free_cell.rs
+++ b/auraed/src/runtime/free_cell.rs
@@ -1,0 +1,27 @@
+use crate::runtime::cell_name::CellName;
+use crate::runtime::CellService;
+use aurae_proto::runtime::{FreeCellRequest, FreeCellResponse};
+use tonic::Status;
+use validation_macros::ValidatedType;
+
+#[derive(ValidatedType)]
+pub(crate) struct ValidatedFreeCellRequest {
+    #[field_type(String)]
+    #[validate]
+    pub cell_name: CellName,
+}
+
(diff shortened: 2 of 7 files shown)


auraed/src/runtime/cell_name.rs
```diff
@@ -0,0 +1,38 @@
...
+            validation::required_not_empty(input, field_name, parent_name)?;
+
+        // TODO: what makes a valid cgroup name
+        // any valid path?
+        // do we want a restriction on length?
+        // anything else?
+        // Highly restrictive for now:
+        validation::allow_regex(
+            &input,
+            &validation::DOMAIN_NAME_LABEL_REGEX,
```
krisnova, line 23: I think is fine. I would encourage us to refuse any `/` character or anything that resembles a filesystem path. 

✅ `my-cgroup-is-awesome`
❌ `my/cgroup/is/awesome`

Mostly because of how Linux creates the cgroup directory structure in `/sys/fs/cgroup`

krisnova (approved):

auraed/src/runtime/cell_name.rs
```diff
@@ -0,0 +1,38 @@
...
+            field_name,
+            parent_name,
+        )?;
+
+        Ok(Self(input))
+    }
+}
+
+impl Deref for CellName {
```
dmah42, line 32: do we need this?  we can always get the str from the String by doing `&*cell_name`.

auraed/src/runtime/free_cell.rs
```diff
@@ -0,0 +1,27 @@
...
+use validation_macros::ValidatedType;
+
+#[derive(ValidatedType)]
+pub(crate) struct ValidatedFreeCellRequest {
+    #[field_type(String)]
+    #[validate]
+    pub cell_name: CellName,
+}
+
+impl FreeCellRequestTypeValidator for FreeCellRequestValidator {}
```
dmah42, line 14: this line makes my head hurt.  i don't know what either of these types are or why we'd need an empty impl of one for the other..

auraed/src/runtime/mod.rs
```diff
@@ -143,12 +148,12 @@ impl cell_service_server::CellService for CellService {
...
-        let cell_name = r.cell_name;
-        info!("CellService: free() cell_name={:?}", cell_name);
-        self.remove_cgroup(&cell_name).expect("remove cgroup");
-        Ok(Response::new(FreeCellResponse {}))
+        let request = request.into_inner();
+        let request = ValidatedFreeCellRequest::validate(request, None)
+            .map_err(|e| Status::invalid_argument(e.to_string()))?;
+
+        info!("CellService: free() cell_name={:?}", &request.cell_name);
+        request.handle(self).map(Response::new)
```
dmah42, line 156: this right here echoes what i wrote above: we're now "calling handle on request passing self" instead of "calling handle on self passing request".  it feels backwards.

auraed/src/runtime/cell_name.rs
future-highway: We do since CellName is not considered a String by the compiler, so `&*cell_name` won't work without implementing  `Deref`

auraed/src/runtime/free_cell.rs
future-highway: This won't always be empty. If the trait `[WHATEVER]TypeValidator` has any required functions, the compiler will complain. In this case, marking the field `#[validate]` tells the macro to generate the default function for validation.

PR #146, opened 2022-12-12
Adding plumbing from stream

krisnova:


api/v0/runtime.proto
```diff
@@ -104,13 +104,28 @@ message Cell {
...
+  //  group get. Note that this is hierarchical, so this is weighted
+  //  against the siblings of this control group.
+  uint64 cpu_shares = 3;
+
+  /// Same syntax as the cpus field of this structure, but applies to
+  //  memory nodes instead of processors.
+  string cpu_mems = 4;
+
+  // In one period, how much can the tasks run in nanoseconds.
+  int64 cpu_quota = 5;
```
dmah42, line 128: we either need a second number to define the period or we need to fix the period and have this be the fraction of that period.  maybe this should be a float32 or something which we translate to microseconds per second or something

auraed/src/runtime/mod.rs
```diff
@@ -292,41 +295,15 @@ impl cell_service_server::CellService for CellService {
...
 fn hierarchy() -> Box<dyn Hierarchy> {
-    hierarchies::auto() // v1/v2 cgroup switch automatically
+    // Auraed will assume the V2 cgroup hierarchy by default.
+    // For now we do not change this, albeit in theory we could
+    // likely create backwards compatability for V1 hierarchy.
+    //
+    // For now, we simply... don't.
+    // hierarchies::auto() // Uncomment to auto detect Cgroup hierarchy
+    // hierarchies::V2
+    Box::new(hierarchies::V2::new())
```
dmah42, line 306: do we need this method? can we not just do this inline?

examples/cells_quota.ts
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env auraescript
```
dmah42, line 1: for christmas (insert festive holiday of your choice) i would like something that executes typescript and checks the cgroup filesystem so we can unit test this stuff :)

dmah42 (approved):

api/v0/runtime.proto
```diff
@@ -104,13 +104,28 @@ message Cell {
   /// See [examples](https://github.com/kata-containers/cgroups-rs/blob/main/tests/builder.rs)
   string name = 1;
 
+  /// A comma-separated list of CPU IDs where the task in the control group
+  // can run. Dashes between numbers indicate ranges.
+  string cpu_cpus = 2;
```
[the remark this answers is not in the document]
krisnova: I don't know if I want to get in the business of doing type mutations at this level "for convenience". 

I think we should pass the fields through 1:1 for right. Maybe in the future once we have the kinks worked out we can debate about changing them or making improvements to how the API is structured.

api/v0/runtime.proto
krisnova: This is just a passthrough to the lower level API for right now. I don't think we can have a second field as it's just a 1:1 passthrough.

auraed/src/runtime/mod.rs
krisnova: I would prefer to have a method that everything calls, such that we can easily change it and add logic in the future. It seems to be safer than just hard coding everything.

api/v0/runtime.proto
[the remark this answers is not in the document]
dmah42: 👍🏻

PR #147, opened 2022-12-12
create cgroup_table as a separate struct so we can add implementation

dmah42:


auraed/src/runtime/mod.rs
```diff
@@ -103,26 +97,25 @@ impl CellService {
...
 
-        let mut cgroup_cache =
-            self.cgroup_table.lock().expect("lock cgroup_table");
-        // Check if there was already a cgroup in the table with this cell name as a key.
-        if let Some(_old_cgroup) =
-            cgroup_cache.insert(cell_name.into(), cgroup.clone())
-        {
-            return Err(anyhow!("cgroup already exists for {cell_name}"));
-        };
+        self.cgroup_table.insert(cell_name, &cgroup).map_err(|e| {
```
krisnova, line 100: What do we do if the cgroup already exists? Should we error here?

dmah42: we do.  it's in the table implementation.

krisnova (approved):

PR #148, opened 2022-12-12
add unit tests to runtime tables

dmah42:
i also added some documentation and tweaked the API so they're consistent and capturing the intent appropriately.


krisnova (approved):
Thank you for adding tests! I really like the `is_error` pattern!

PR #149, opened 2022-12-12
make clean clean

dmah42:
make clean should fully clean stuff but doesn't.  now it does.


dmah42:
nevermind, it seems that cargo clean should do the right thing.

PR #150, opened 2022-12-13
CellService validation

future-highway:
This integrates validation for `CellService`.

Things to notice:
- The core logic now exists on `impl CellService`, not `impl cell_service_server::CellService for CellService`; the trait implementation is all boilerplate.
- All the functions that are correlated to rpc endpoints are passed the request message post validation.

Since, post merging this, we will need to work with the validated types, it may get in the way (i.e., slow down) developing the core logic. For example, if adding a field to a proto message, a good temporary solution is to also add the field to the validated struct, decorate it with `#[validate(none)]`, and leave a todo. For example:

```proto
message AllocateCellRequest {
  ...
  uint64 my_new_field = 10;
}
```    

```rust
pub(crate) struct ValidatedAllocateCellRequest {
  ...
  // TODO: Need to actually validate
  #[validate(none)]
  pub my_new_field: u64
}
```


krisnova (approved):
Still looks good to me! I think my main question is how easy is it to "do the plumbing" with these? If I wanted to introduce another service to Aurae how much time would we spend plumbing through new validation?

future-highway:
That's a fair question and one that echos concerns dma had also. We settled at, we should feel free to avoid caring about the plumbing while initially forming the API as it is likely going to be changing rapidly. Once the API starts to firm up, it's not difficult to refactor.

Personally, I think/hope as the plumbing is used (along with stronger typing, like `CellName`, `ExecutableName`) its benefits will become more apparent, and show how it actually helps in writing core logic (at least that's how I feel).

auraed/src/runtime/cell_name.rs
```diff
@@ -25,6 +30,23 @@ impl validation::ValidatedField<String> for CellName {
...
+
+impl ValidatedField<String> for CellName {
+    fn validate(
+        input: Option<String>,
+        field_name: &str,
+        parent_name: Option<&str>,
+    ) -> Result<Self, ValidationError> {
+        let input =
+            validation::required_not_empty(input, field_name, parent_name)?;
+
```
dmah42, line 49: is there any way to ensure the validation doesn't mutate the input? I can't think of anything in the type system we could use but it could be a neat constraint to have.  in C++ I'd do something with const and not returning, ie just having read access to the thing under validation.

auraed/src/runtime/cgroup_table.rs
```diff
@@ -24,17 +24,16 @@ impl CgroupTable {
...
     ) -> Result<()> {
         let mut cache = self
             .cache
             .lock()
             .map_err(|e| anyhow!("failed to lock cgroup_table: {e:?}"))?;
         // Check if there was already a cgroup in the table with this cell name as a key.
-        if let Some(_old_cgroup) =
-            cache.insert(cell_name.into(), cgroup.clone())
-        {
+        // TODO: This seems like a bug since it potentially replaces an existing value. Guard against this scenario.
```
dmah42, line 35: I thought the insert would fail. good catch. I guess we need `entry(...).or_insert` or something.

auraed/src/runtime/cgroup_table.rs
```diff
@@ -24,17 +24,16 @@ impl CgroupTable {
     /// Returns an error if a duplicate [cell_name] already exists in the cache.
     pub(crate) fn insert(
         &self,
-        cell_name: &str,
-        cgroup: &Cgroup,
+        cell_name: String,
```
dmah42, line 27: why this type change? we shouldn't be taking ownership of the cell name so I think this should still be a borrow (from which we can clone) no?

auraed/src/runtime/executable_name.rs
```diff
@@ -0,0 +1,61 @@
+use std::fmt::{Display, Formatter};
+use std::ops::Deref;
+use validation::{ValidatedField, ValidationError};
+
+#[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
+pub(crate) struct ExecutableName(String);
+
+impl ExecutableName {
+    pub fn validate_for_creation(
+        input: Option<String>,
```
dmah42, line 10: why would we ever pass None to validation?

dmah42 (approved):
really like the pattern of keeping the service layer light (validate and call impl) :)

auraed/src/runtime/cell_name.rs
future-highway: I'm assuming you mean the functions in the validation crate:

I just double checked, and only the functions starting with `required` require ownership (`valid_json` does too, but that is a mistake and unnecessary). On the `required` functions, it makes for a nicer pattern for the caller as it will either return an error or unwrap the `Option`, but it is only designed this way for convenience, and can be changed.  

We may also be able to restrict the generic functions further to prevent passing ownership.

auraed/src/runtime/cgroup_table.rs
future-highway: Yeah, the docs say it will update.

auraed/src/runtime/cgroup_table.rs
future-highway: I consider taking a reference and then immediately cloning to be "lying" to the caller. Insert needs ownership of `cell_name`, so the signature should represent that. It may also be a scenario where the caller does not need ownership and we can therefore avoid a clone.

auraed/src/runtime/executable_name.rs
future-highway: Obviously we shouldn't, or validation would fail, but the function doesn't know that you will always have a `String`. Essentially, it is validating that you even have a value. Given the way the api/protos are currently, we wouldn't have that scenario.

However, I feel there are benefits to keeping a consistent signature style for the `validate_for_creation` functions. I probably should have made it a function on the `ValidatedField` trait, with a default implementation. I'll do that.

auraed/src/runtime/mod.rs
```diff
@@ -237,29 +192,128 @@ impl cell_service_server::CellService for CellService {
...
             "Child process with pid {child_id} exited with status {exit_status}",
         );
 
-        // Ok
         Ok(Response::new(StopCellResponse {}))
     }
+
+    // Here is where we define the "default" cgroup parameters for Aurae cells
+    fn create_cgroup(&self, cell: ValidatedCell) -> Result<Cgroup, Error> {
+        let ValidatedCell {
```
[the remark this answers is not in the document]
future-highway: I'm not sure I am understanding this question, so I'll answer it a couple of ways:

If the concern is the destructuring of `ValidatedCell` (i.e., `let ValidatdCell { all, the, props, is, a, lot } = cell;`) and typing out all those props, I (using clion, but I think it works in VSCode too) don't actually type any of the fields. All I type is `let ValidatedCell {} = cell;` and then the IDEs quick fix takes care of the rest. I know that's not helpful if someone uses an IDE that doesn't do it for them, but there are other benefits to destructuring. Specifically, in this function, I was able to pass the value of each field into `CgroupBuilder` without needing to clone anything.

If you are asking about converting a `Cell` type to a `ValidatedCell` type, that happens in the validation step. Most of it is generated. If you expand the `ValidatedType` macro attribute on top of `ValidatedCell` you get:

```rust
impl ::validation::ValidatedType<Cell> for ValidatedCell {
    fn validate(input: Cell, parent_name: Option<&str> ) -> ::std::result::Result<ValidatedCell, ::validation::ValidationError> {
        CellValidator::validate_fields(&input, parent_name)?;
        let Cell { name, cpu_cpus, cpu_shares, cpu_mems, cpu_quota, } = input;
        let name = CellValidator::validate_name(name, "name", parent_name)?;
        let cpu_cpus = CellValidator::validate_cpu_cpus(cpu_cpus, "cpu_cpus", parent_name)?;
        let cpu_shares = CellValidator::validate_cpu_shares(cpu_shares, "cpu_shares", parent_name)?;
        let cpu_mems = CellValidator::validate_cpu_mems(cpu_mems, "cpu_mems", parent_name)?;
        let cpu_quota = CellValidator::validate_cpu_quota(cpu_quota, "cpu_quota", parent_name)?;
        Ok(ValidatedCell { name, cpu_cpus, cpu_shares, cpu_mems, cpu_quota })
    }
}
trait CellTypeValidator {
    fn validate_name(name: String, field_name: &str, parent_name: Option<&str>) -> ::std::result::Result<CellName, ::validation::ValidationError>;
    fn validate_cpu_cpus(cpu_cpus: String, field_name: &str, parent_name: Option<&str>) -> ::std::result::Result<String, ::validation::ValidationError> { Ok(cpu_cpus) }
    fn validate_cpu_shares(cpu_shares: u64, field_name: &str, parent_name: Option<&str>) -> ::std::result::Result<u64, ::validation::ValidationError> { Ok(cpu_shares) }
    fn validate_cpu_mems(cpu_mems: String, field_name: &str, parent_name: Option<&str>) -> ::std::result::Result<String, ::validation::ValidationError> { Ok(cpu_mems) }
    fn validate_cpu_quota(cpu_quota: i64, field_name: &str, parent_name: Option<&str>) -> ::std::result::Result<i64, ::validation::ValidationError> { Ok(cpu_quota) }
    fn validate_fields(fields: &Cell, parent_name: Option<&str>) -> ::std::result::Result<(), ::validation::ValidationError> { Ok(()) }
}
struct CellValidator;
```  

The above code does:
1. Generate a `validate` function that destructures `Cell` and calls a validation function for each field.
2. The field validation functions are generated next. In this case the `CellTypeValidator` trait. Because I decorated most of the fields with `#[validate(none)]` most of them have pass through default implementations. `validate_name` is the only function that requires an implementation. If I had added `#[validate]` to the `name` field, I would get `validation::ValidatedField::validate(Some(name), field_name, parent_name)` which is the same as calling `CellName::validate`. There is also `#[validate(opt)]` which would call `CellName::validate_optional`.
3. It creates an empty struct to implement the `TypeValidator` trait on. It needs to create it to have control over the name of the struct so that it knows how to call it. 

An additional note, if all the fields are able to to be decorated with either `#[validate]`, `#[validate(opt)]`, or `#[validate(none)]` the `TypeValidator` trait would not have any functions requiring implementations, which lead to `impl CellTypeValidator for CellValidator {}` which I know you noted before as being a little weird, but it's just the result of the macro taking care of everything else.  Perhaps I should consider having the macro also generate that last bit in the event all the fields validation functions have default implementations.

PR #152, opened 2022-12-13
Validation updates

future-highway:
Moved `validate_for_creation` into the `ValidatedField` trait, and expanded the macro capabilities to allow generating the call to it automatically.


dmah42 (approved):
this is one of those grumpy approvals because it means i'll have to rework my in-flight PR :P

PR #153, opened 2022-12-13
introduce fixed period for cpu quota

dmah42:
in cgroups, cpu quota is defined as a portion of time within a defined period. with this PR we define a period to be 1 second (in microseconds) and then allow the cpu quota to be set as a fraction of that (in microseconds).


auraed/src/runtime/validation.rs
```diff
@@ -88,8 +89,8 @@ pub(crate) struct ValidatedCell {
     pub cpu_shares: u64,
     #[validate(none)]
     pub cpu_mems: String,
-    #[validate(none)]
-    pub cpu_quota: i64,
+    #[field_type(i64)]
+    pub cpu_quota: CpuQuota,
```
future-highway, line 93: The error is here. You need to add `#[validate(create)]` to automatically call `CpuQuota::validate_for_creation`

auraed/src/runtime/cpu_quota.rs
```diff
@@ -0,0 +1,62 @@
...
+
+    fn validate_for_creation(
+        input: Option<i64>,
+        field_name: &str,
+        parent_name: Option<&str>,
+    ) -> Result<Self, ValidationError> {
+        let input = Self::validate(input, field_name, parent_name)?;
+
+        validation::maximum_value(
+            *input,
```
future-highway, line 40: I don't think this should need a deref. Not a big deal

future-highway: Nevermind. It does, lol

future-highway: Threw me off because I would usually do `input.0` in this case.

auraed/src/runtime/mod.rs
```diff
@@ -236,7 +238,8 @@ impl CellService {
             .cpu()
             .shares(cpu_shares)
             .mems(cpu_mems)
-            .quota(cpu_quota)
+            .period(1000000) // microseconds in a second
+            .quota(*cpu_quota)
```
future-highway, line 242: I think this copies the i64, which is cheap, but we could use `cpu_quota.into_inner()` to avoid the copy.

future-highway (approved):

future-highway (approved):

PR #156, opened 2022-12-13
clean up some dependencies

dmah42:
* remove unused deps using `cargo machete --with-metadata`
* move common dependencies to the workspace to ensure consistent
versioning

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,33 @@
+# ---------------------------------------------------------------------------- #
+#             Apache 2.0 License Copyright © 2022 The Aurae Authors            #
+#                                                                              #
+#                +--------------------------------------------+                #
+#                |   █████╗ ██╗   ██╗██████╗  █████╗ ███████╗ |                #
+#                |  ██╔══██╗██║   ██║██╔══██╗██╔══██╗██╔════╝ |                #
+#                |  ███████║██║   ██║██████╔╝███████║█████╗   |                #
+#                |  ██╔══██║██║   ██║██╔══██╗██╔══██║██╔══╝   |                #
+#                |  ██║  ██║╚██████╔╝██║  ██║██║  ██║███████╗ |                #
+#                |  ╚═╝  ╚═╝ ╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═╝╚══════╝ |                #
+#                +--------------------------------------------+                #
+#                                                                              #
+#                         Distributed Systems Runtime                          #
+#                                                                              #
+# ---------------------------------------------------------------------------- #
+#                                                                              #
+#   Licensed under the Apache License, Version 2.0 (the "License");            #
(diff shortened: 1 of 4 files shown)


auraed/Cargo.toml
```diff
@@ -33,42 +33,34 @@ name = "auraed"
...
+tokio = { workspace = true, features = ["macros", "fs", "rt-multi-thread"] }
+tokio-stream = { version = "0.1", features = ["net", "sync"] }
+tonic = { workspace = true, features = ["tls"] }
 validation = { path = "../crates/validation", features=["regex"] }
 validation_macros = { path = "../crates/validation/macros" }
-
-[build-dependencies]
-anyhow = "1.0.65"
-tonic-build = "0.8"
+walkdir = "2"
```
future-highway, line 63: This doesn't seem right. The compiler is probably allowing it because we are using it in what is temporarily dead code: https://github.com/aurae-runtime/aurae/blob/dfaacc1d5e583965aa527a192654de913a29c329/auraed/src/init/fileio.rs#L34

auraed/Cargo.toml
```diff
@@ -33,42 +33,34 @@ name = "auraed"
...
+license = "Apache-2.0"
 
 [[bin]]
 name = "auraed"
 path = "src/bin/main.rs"
 
 [dependencies]
+anyhow = {workspace = true}
+aurae-proto = { workspace = true }
+cgroups-rs = { git = "https://github.com/kata-containers/cgroups-rs", branch = "main"}
```
future-highway, line 45: Wondering why we aren't using the crate from crates.io

Cargo.toml
```diff
@@ -7,3 +37,9 @@ members = [
     "crates/validation",
     "crates/validation/macros"
 ]
+
+[workspace.dependencies]
+anyhow = "1.0.66"
+aurae-proto = {path = "./aurae-proto", version = "0.1.0"}
```
future-highway, line 43: Not sure there is any effect from `version` when using `path`

future-highway (changes requested):
Just the walkdir comment.

auraed/Cargo.toml
dmah42: we are.. i haven't removed this.

auraed/Cargo.toml
dmah42: i haven't removed this either.

Cargo.toml
dmah42: probably not.. i'm just copying what was there before.

auraed/Cargo.toml
future-highway: Nope. My mistake. Saw `[build-dependencies]` and thought you moved it there. It didn't click in my brain that that line is not going to be there.

future-highway (approved):

PR #159, opened 2022-12-14
better error handling in the runtime

dmah42:
using thiserror more properly enables the runtime mod.rs
functionality to have less noise.

### DIFF
--- a/auraed/src/runtime/error.rs
+++ b/auraed/src/runtime/error.rs
@@ -27,17 +27,17 @@ pub(crate) enum RuntimeError {
 
 impl From<RuntimeError> for Status {
     fn from(err: RuntimeError) -> Self {
-        let err_msg = err.to_string();
-        error!("{err_msg}");
+        let msg = err.to_string();
+        error!("{msg}");
         match err {
             RuntimeError::MissingArgument { arg: _ }
             | RuntimeError::ValidationError { 0: _ } => {
-                Self::failed_precondition(err_msg)
+                Self::failed_precondition(msg)
             }
             RuntimeError::Internal { msg: _, err: _ }
             | RuntimeError::CgroupsError { 0: _ }
-            | RuntimeError::Other { 0: _ } => Self::internal(err_msg),
-            RuntimeError::Unallocated { resource } => Self::not_found(msg),
+            | RuntimeError::Other { 0: _ } => Self::internal(msg),
+            RuntimeError::Unallocated { resource: _ } => Self::not_found(msg),
         }
     }
 }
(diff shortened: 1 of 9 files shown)


auraed/src/runtime/error.rs
```diff
@@ -2,27 +2,40 @@ use log::error;
...
 use tonic::Status;
 
-#[derive(Error, Debug, PartialEq, Eq)]
-pub enum CellServiceError {
+pub(crate) type Result<T> = std::result::Result<T, RuntimeError>;
+
+#[derive(Error, Debug)]
+pub(crate) enum RuntimeError {
     #[error("missing argument in request: {arg}")]
     MissingArgument { arg: String },
```
future-highway, line 10: We can take this out as validation takes care of it.

auraed/src/runtime/error.rs
```diff
@@ -2,27 +2,40 @@ use log::error;
...
+    ValidationError(#[from] validation::ValidationError),
+
     #[error("internal error: {msg}: {err}")]
     Internal { msg: String, err: String },
+
+    #[error(transparent)]
+    CgroupsError(#[from] cgroups_rs::error::Error),
+
+    #[error(transparent)]
+    Other(#[from] anyhow::Error),
```
future-highway, line 22: Ideally we will eventually eliminate anyhow (I think)

auraed/src/runtime/error.rs
```diff
@@ -2,27 +2,40 @@ use log::error;
...
 }
 
-impl From<CellServiceError> for Status {
-    fn from(err: CellServiceError) -> Self {
+impl From<RuntimeError> for Status {
```
future-highway, line 25: Looks like you were a step ahead of me, lol

auraed/src/runtime/error.rs
dmah42: i tried, but i needed this to go from ValidationError to RuntimeError to Status.

i didn't see anything that would convert ValidationError straight to Status.. did i miss it?

auraed/src/runtime/error.rs
dmah42: i agree.. though it's convenient for creating these quick errors from a string.  i think if we stick to `anyhow!` _only_ we might be ok.

auraed/src/runtime/error.rs
dmah42: :D

future-highway (dismissed):

future-highway (approved):

PR #161, opened 2022-12-14
Accept multiple executables in StartCellRequest

future-highway:
Also shows how to format `parent_name` for `Vec` so that it matches the grpc conventions (but maybe there is a more efficient way?)


auraed/src/runtime/mod.rs
```diff
@@ -117,40 +117,44 @@ impl CellService {
...
+            let cgroup =
+                self.cgroup_table.get(&cell_name)?.ok_or_else(|| {
+                    RuntimeError::Unallocated { resource: "cgroup".into() }
+                })?;
+
+            // Add the newly started child process to the cgroup
+            let cgroup_pid = CgroupPid::from(child.id() as u64);
+            cgroup.add_task(cgroup_pid).map_err(RuntimeError::from)?;
+
+            info!("CellService: spawn() -> pid={:?}", &child.id());
```
dmah42, line 154: we should log cell name here too for when we are handling multiple requests

dmah42 (approved):

dmah42 (approved):

PR #163, opened 2022-12-14
WIP initial attempt at a CellTable

dmah42:
** does not pass borrow checker **

### DIFF
--- a/auraed/src/runtime/cell_table.rs
+++ b/auraed/src/runtime/cell_table.rs
@@ -1,0 +1,112 @@
+use std::{
+    collections::HashMap,
+    process::Child,
+    sync::{Arc, Mutex},
+};
+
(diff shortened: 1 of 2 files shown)


krisnova:
Things to think about: 

 - We will likely need to print this structure to stdout using an Aurae client in the future.
 - Do we want `CellTable` to have `add_cgroup` or `add_cell`?
 - At what point does a `Cell` stop being a cell and start being a `cgroup`?

For what it's worth in my opinion a cell is just "A managed cgroup with special pre-exec functionality" which seems like a good balance and a good way of defining it.

As Aurae grows our pre-exec functionality is likely to grow, so I think drawing that distinction now is helpful.

dmah42:
this was superceded by @future-highway's work.

PR #165, opened 2022-12-14
Introduce cells module

future-highway:


auraed/src/cells/executable.rs
```diff
@@ -0,0 +1,26 @@
+use cgroups_rs::CgroupPid;
+use log::info;
+use std::io;
+use std::process::{Child, Command, ExitStatus};
+
+#[derive(Debug)]
+pub(crate) struct Executable(Child);
+
+impl Executable {
+    pub fn spawn(mut command: Command) -> io::Result<Self> {
```
dmah42, line 10: nit: i still think it's weird that there's no args passed through separately :)

auraed/src/cells/mod.rs
```diff
@@ -0,0 +1,24 @@
+pub(crate) use cell::Cell;
```
dmah42, line 1: i'm finding it weird that "cells" is a crate outside "runtime" as it only exists as a concept inside runtime.

does it need to be a separate crate?

auraed/src/runtime/error.rs
```diff
@@ -29,14 +28,13 @@ impl From<RuntimeError> for Status {
...
-                //| RuntimeError::ValidationError { 0: _ } => {
+            RuntimeError::MissingArgument { .. } => {
                 Self::failed_precondition(msg)
             }
-            RuntimeError::Internal { msg: _, err: _ }
-            | RuntimeError::CgroupsError { 0: _ }
-            | RuntimeError::Other { 0: _ } => Self::internal(msg),
-            RuntimeError::Unallocated { resource: _ } => Self::not_found(msg),
+            RuntimeError::Internal { .. }
+            | RuntimeError::CellsError { .. }
```
dmah42, line 35: is it true that all CellsErrors are internal?  i expected that we might have different outer "public" errors coming from different methods within each submodule so having this top-level runtime error was the right way to determine the mapping from one set of errors to the statuses.

auraed/src/cells/mod.rs
future-highway: We can nest it under runtime. Or just drag the files to runtime. One benefit of small crates to keep in mind is build caching is done per crate.

auraed/src/runtime/mod.rs
```diff
@@ -28,95 +28,85 @@
...
         &self,
         request: ValidatedAllocateCellRequest,
-    ) -> std::result::Result<Response<AllocateCellResponse>, Status> {
+    ) -> Result<Response<AllocateCellResponse>, Status> {
         // Initialize the cell
         let ValidatedAllocateCellRequest { cell } = request;
-
         info!("CellService: allocate() cell={:?}", cell);
 
+        if self.cells.contains(&cell.name).map_err(RuntimeError::from)? {
```
dmah42, line 74: this could be an error on cells.insert, no?

dmah42 (commented):
super clean.. just some nits and requests for clarification :)

auraed/src/runtime/error.rs
future-highway: I expect you are right. We will need to tackle mapping them. My suggestion would be to define the errors better (i.e., CellExits, CellNotFound, etc.) and then that would be easier to map to Status correctly

auraed/src/runtime/mod.rs
future-highway: Yes, but then you would have already called 'Cell::allocate` (ln 84)

future-highway: It checks on insert too, I think

auraed/src/cells/executable.rs
future-highway: I think you were 100% right on not having it in validation. Currently it's in `CellService`. I'm inclined to continue with your suggestion to keep passing it down, especially since I forgot to pass description.

krisnova (dismissed):
3 things came to mind that are likely out of scope for this work. 

 - We should change `aurae_process_pre_exec` to a better name
 - The unit tests are commented out
 - The `info!()` statements should either be removed, made dynamic, or standardized for all methods in `impl Cell`

dmah42:
> 3 things came to mind that are likely out of scope for this work.
> 
> * We should change `aurae_process_pre_exec` to a better name

+1

> * The unit tests are commented out

i think these don't pass as they need to fiddle with the file system and github doesn't like it... but i haven't tried to debug it.

> * The `info!()` statements should either be removed, made dynamic, or standardized for all methods in `impl Cell`

i don't know what you mean by dynamic but i think they're helpful to have to see the service working.

krisnova (approved):

auraed/src/cells/mod.rs
dmah42: I think we should keep it nearer the cell service. it provides clearer visibility expectations.

krisnova:
My only gripe is source code headers that we need to include for #168 

Not required for this PR but in general we should look at including headers for new files or bulk removing them from the project.

dmah42 (approved):

dmah42 (approved):

dmah42 (approved):

PR #167, opened 2022-12-15
introduced cpuset_mems

dmah42:


krisnova:
What is our policy on adding docs for the fields?

krisnova (dismissed):

dmah42:
> What is our policy on adding docs for the fields?

i have been working under the assumption that the api proto fields are the ones that need documentation but i'm not sure how much to link out to other docs vs repeat information.  open to suggestions.

future-highway (approved):

PR #169, opened 2022-12-15
Modeling cells module errors

future-highway:
Made a couple of decisions based on assumptions:

1. We don't want to expose source errors (e.g., io::Error) that reveal the workings of the code, but that seems like it may not be a concern in this project.
2. Splitting errors into smaller focused errors that then feed upward (like I did in `init`) is a good idea. Generally I like it, but the recursiveness of `Cell` is causing me to question that approach in this case.

Overall, part of the goal was to use the language of aurae. For example, "start executable" instead of "spawn command/process/child"

### DIFF
--- a/auraed/src/runtime/cells/cell.rs
+++ b/auraed/src/runtime/cells/cell.rs
@@ -186,7 +186,7 @@ pub(crate) enum CellError {
     ExecutableNotFound { cell_name: CellName, executable_name: ExecutableName },
     #[error("cell '{cell_name}': {source}")]
     ExecutableError { cell_name: CellName, source: ExecutableError },
-    #[error("cell '{cell_name}' failed to add executable (executable:?)")]
+    #[error("cell '{cell_name}' failed to add executable (executable:?): {source}")]
     FailedToAddExecutable {
         cell_name: CellName,
         executable: Executable,
@@ -203,13 +203,9 @@ impl From<CellError> for Status {
             | CellError::ExecutableExists { .. } => Status::already_exists(msg),
             CellError::CellNotFound { .. }
             | CellError::ExecutableNotFound { .. } => Status::not_found(msg),
-            // TODO (future-highway): I don't know what the conventions are of revealing
-            //  messages that reveal the workings of the system to the api consumer
-            //  in this type of application.
-            //  For now, taking the safe route and not exposing the error messages for the below errors.
             CellError::ExecutableError { .. }
             | CellError::FailedToFreeCell { .. }
-            | CellError::FailedToAddExecutable { .. } => Status::internal(""),
+            | CellError::FailedToAddExecutable { .. } => Status::internal(msg),
         }
     }
 }
--- a/auraed/src/runtime/cells/cell.rs
+++ b/auraed/src/runtime/cells/cell.rs
@@ -1,8 +1,37 @@
+/* -------------------------------------------------------------------------- *\
+ *             Apache 2.0 License Copyright © 2022 The Aurae Authors          *
+ *                                                                            *
+ *                +--------------------------------------------+              *
+ *                |   █████╗ ██╗   ██╗██████╗  █████╗ ███████╗ |              *
+ *                |  ██╔══██╗██║   ██║██╔══██╗██╔══██╗██╔════╝ |              *
+ *                |  ███████║██║   ██║██████╔╝███████║█████╗   |              *
+ *                |  ██╔══██║██║   ██║██╔══██╗██╔══██║██╔══╝   |              *
+ *                |  ██║  ██║╚██████╔╝██║  ██║██║  ██║███████╗ |              *
+ *                |  ╚═╝  ╚═╝ ╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═╝╚══════╝ |              *
+ *                +--------------------------------------------+              *
+ *                                                                            *
+ *                         Distributed Systems Runtime                        *
+ *                                                                            *
+ * -------------------------------------------------------------------------- *
+ *                                                                            *
+ *   Licensed under the Apache License, Version 2.0 (the "License");          *
+ *   you may not use this file except in compliance with the License.         *
+ *   You may obtain a copy of the License at                                  *
+ *                                                                            *
+ *       http://www.apache.org/licenses/LICENSE-2.0                           *
+ *                                                                            *
+ *   Unless required by applicable law or agreed to in writing, software      *
+ *   distributed under the License is distributed on an "AS IS" BASIS,        *
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ *   See the License for the specific language governing permissions and      *
+ *   limitations under the License.                                           *
+ *                                                                            *
+\* -------------------------------------------------------------------------- */
+
 use super::Result;
 use crate::runtime::cells::{
     validation::ValidatedCell, CellName, CellsError, Executable, ExecutableName,
 };
-use anyhow::anyhow;
 use cgroups_rs::cgroup_builder::CgroupBuilder;
 use cgroups_rs::{hierarchies, Cgroup, Hierarchy};
 use log::info;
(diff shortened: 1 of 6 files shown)


auraed/src/runtime/cells/cell.rs
```diff
@@ -41,25 +72,28 @@ impl Cell {
         Self { name, cgroup, executables: Default::default() }
     }
 
-    pub fn free(self) -> Result<()> {
-        self.cgroup.delete()?;
+    pub fn free(self) -> Result<(), CellError> {
```
dmah42, line 75: `type Result<T> = Result<T, CellError>` can simplify some of the code here.. at least it makes it easy to spot when we're being inconsistent (if we are).

auraed/src/runtime/cells/cell.rs
```diff
@@ -41,25 +72,28 @@ impl Cell {
...
-                self.name
-            )
-            .into());
+            return Err(CellError::ExecutableExists {
+                cell_name: self.name.clone(),
+                executable_name: exe_name,
+            });
         }
 
         let _ = command.args(args);
```
dmah42, line 99: this assumes that command.args doesn't return a new instance of a Command.  is that assuming too much about the implementation?

auraed/src/runtime/cells/cell.rs
```diff
@@ -74,11 +108,26 @@ impl Cell {
         };
 
         // Start the child process
-        let exe = Executable::spawn(command)?;
+        let exe =
+            Executable::start(exe_name.clone(), command).map_err(|e| {
```
dmah42, line 112: why wouldn't Executable::start (part of the cells mod) return an ExecutableError directly (and avoid the map)?

auraed/src/runtime/cells/cell.rs
```diff
@@ -74,11 +108,26 @@ impl Cell {
...
+                CellError::ExecutableError {
+                    cell_name: self.name.clone(),
+                    source: e,
+                }
+            })?;
 
         // Add the newly started child process to the cgroup
         let exe_pid = exe.pid();
-        self.cgroup.add_task(exe.pid()).map_err(CellsError::from)?;
+        match self.cgroup.add_task(exe.pid()) {
```
dmah42, line 121:
```
if let Err(e) = self.cgroup.add_task(exe.pid()) {
  return Err(...);
};
```

is easier i think?

auraed/src/runtime/cells/cell.rs
```diff
@@ -91,17 +140,28 @@ impl Cell {
...
-    ) -> Result<ExitStatus> {
-        if let Some(exe) = self.executables.remove(exe_name) {
-            Ok(exe.kill()?)
+    ) -> Result<ExitStatus, CellError> {
+        if let Some(mut exe) = self.executables.remove(exe_name) {
+            match exe.kill() {
+                Ok(exit_status) => Ok(exit_status),
+                Err(e) => {
+                    // Failed to kill, put it back in cache
+                    let _ = self.executables.insert(exe_name.clone(), exe);
```
dmah42, line 152: instead of remove/fail/insert.. how about get/kill/remove?

ie get it, try to kill it, remove it if successful

auraed/src/runtime/cells/cell.rs
```diff
@@ -129,3 +189,48 @@ fn hierarchy() -> Box<dyn Hierarchy> {
     // hierarchies::V2
     Box::new(hierarchies::V2::new())
 }
+
+#[derive(Error, Debug)]
+pub(crate) enum CellError {
+    #[error("cell '{cell_name}' already exists'")]
+    Exists { cell_name: CellName },
```
dmah42, line 196: nit: "CellExists" to parallel ExecutableExists ?

auraed/src/runtime/cells/cell.rs
```diff
@@ -129,3 +189,48 @@ fn hierarchy() -> Box<dyn Hierarchy> {
...
+        match err {
+            CellError::Exists { .. } | CellError::ExecutableExists { .. } => {
+                Status::already_exists(msg)
+            }
+            CellError::NotFound { .. }
+            | CellError::ExecutableNotFound { .. } => Status::not_found(msg),
+            // TODO (future-highway): I don't know what the conventions are of revealing
+            //  messages that reveal the workings of the system to the api consumer
+            //  in this type of application.
+            //  For now, taking the safe route and not exposing the error messages for the below errors.
```
dmah42, line 230: this won't expose it to the client, but the auraed will still log it above (l:220).  up to you how much you want to provide to the client vs only have in the server logs :)

auraed/src/runtime/cells/cell_service.rs
```diff
@@ -63,11 +63,7 @@ impl CellService {
         info!("CellService: allocate() cell={:?}", cell);
 
         if self.cells.contains(&cell.name)? {
-            return Err(CellsError::Other(anyhow!(
-                "cell '{}' already exists",
-                cell.name
-            ))
-            .into());
+            return Err(CellError::Exists { cell_name: cell.name }.into());
```
dmah42, line 66: it frustrates me that this is here and not an error on insert, but i know we already "allocate" the cell so we need to check it first.  should allocate do this check anyway? otherwise we'll be trying to create a duplicate fs directory for this cgroup i think.

auraed/src/runtime/cells/cell.rs
future-highway: Yup, that's a great point. This can break tomorrow and we wouldn't know

auraed/src/runtime/cells/executable.rs
```diff
@@ -1,26 +1,76 @@
...
+                source: e,
+            }),
+        }
     }
 
-    pub fn kill(mut self) -> io::Result<ExitStatus> {
-        let id = self.0.id();
-        self.0.kill()?;
-        let exit_status = self.0.wait()?;
+    pub fn kill(&mut self) -> Result<ExitStatus, ExecutableError> {
```
dmah42, line 50: should this be "stop" now?

auraed/src/runtime/cells/cell.rs
future-highway: Will add it

auraed/src/runtime/cells/cell.rs
future-highway: Will update

auraed/src/runtime/cells/cell.rs
future-highway: I'm not sure my comment was clear enough. So...

For example, the `CellError::FailedToFree` variant has a `source: cgroups_rs::error::Error` field and an error template of `"cell '{cell_name}' could not be freed: {source}"`, so source ends up in the error message, which I was concerned about ending up in the `Status`'s message that is returned from `CellService`.

auraed/src/runtime/cells/cell.rs
future-highway: I'm torn. Will go with your suggestion.

auraed/src/runtime/cells/cell.rs
dmah42: sure.. is it going to be the same person seeing the Status Response as is running auraed (in which it's logged to stderr)?

auraed/src/runtime/cells/cell.rs
future-highway: The executable doesn't know what cell it is in. On the map, the cell wraps the executable error in its own (`CellError`) and adds the name of the cell.

Lets not talk about recursion just yet, lol

auraed/src/runtime/cells/executable.rs
future-highway: I was going with the assumption that there are a couple of signals we may want to send the process, and that the api would be expanded to accommodate that. 

I was initially going to make this fn private and add `pub fn stop(signal: SignalsThatStopTheExeEnum)`, which would then call the private implementations for each signal, but I didn't feel this PR was the one to tackle that in (aka I couldn't do it in the 2 minutes I gave it before getting  back to the error stuff for the PR, lol).

auraed/src/runtime/cells/cell_service.rs
future-highway: I'm not sure I understand. We are in `CellService::allocate` here within the check for `self.cells.contains`, but before we call `Cell::allocate`.

Do you mean `Cell::allocate` should check with the system directly? Probably not a bad idea, but if we get there, our cache has become inconsistent with the system (which is something I agree we need to consider a possibility and try to heal).

auraed/src/runtime/cells/cell.rs
future-highway: Sounds like you're saying it's fine to expose, which makes perfect sense considering the observability api. I'm still training my brain to ignore the lessons ingrained from writing other programs.

dmah42 (approved):

PR #170, opened 2022-12-16
Organizing in cells module

future-highway:
This is mostly just a few changes moving things around and being a little opinionated on how the code base should be, to hopefully help keep things consistent and neat.

Feel free to accept or reject the whole thing or pick and choose (which I can put in a clean PR).


auraed/src/runtime/cells/cell.rs
```diff
@@ -28,19 +28,62 @@
...
-use crate::runtime::cells::{
-    validation::ValidatedCell, CellName, Executable, ExecutableName,
+use super::{
+    validation::ValidatedCell, CellName, Executable, ExecutableError,
+    ExecutableName,
 };
 use cgroups_rs::cgroup_builder::CgroupBuilder;
 use cgroups_rs::{hierarchies, Cgroup, Hierarchy};
-use log::{error, info};
+use log::error;
```
dmah42, line 37: is there a reason to split these "use" statements?

auraed/src/runtime/cells/cell.rs
```diff
@@ -28,19 +28,62 @@
...
+use log::info;
 use std::collections::HashMap;
 use std::process::{Command, ExitStatus};
 use thiserror::Error;
 use tonic::Status;
 
-type Result<T> = std::result::Result<T, CellError>;
+pub(crate) type Result<T> = std::result::Result<T, CellError>;
```
dmah42, line 44: i prefer to keep this as a separate file, only because it keeps this file focused on the Cell functionality.  you have it at `pub(crate)` visibility anyway so i don't think it matters?

auraed/src/runtime/cells/cell_service.rs
```diff
@@ -28,22 +28,43 @@
...
+    #[error(transparent)]
+    CellError(#[from] CellError),
+    #[error("failed to lock cells table")]
+    FailedToObtainLock(),
+}
+
+impl From<CellServiceError> for Status {
+    fn from(err: CellServiceError) -> Self {
+        match err {
+            CellServiceError::CellError(err) => err.into(),
```
dmah42, line 60: this is still weird to me that we have one module with two error types, one of which is just translated into the other... feels like we should have one at each "service" level.

auraed/src/runtime/cells/cell_service.rs
```diff
@@ -57,7 +78,7 @@ impl CellService {
     fn allocate(
         &self,
         request: ValidatedAllocateCellRequest,
-    ) -> Result<Response<AllocateCellResponse>, Status> {
+    ) -> Result<AllocateCellResponse> {
```
dmah42, line 81: this is great

auraed/src/runtime/cells/cell.rs
future-highway: Assuming you mean for the `Result` and `CellError`:

I would argue that `CellError` and therefore also `Result` is part of the cell functionality as it represents what could go wrong when using `Cell`, and is likely to change with changes to `Cell`. Maybe I should put `CellError` back at the bottom?

future-highway: I moved the error types to the bottom of the files. See what you think

dmah42: i think i'm stuck on "cells" or "the cell service" should have a single error type and that means it shouldn't be in this file.  however, if you separate the "cell service error" from the "cell error" then i can see why you'd keep the CellError here.

auraed/src/runtime/cells/cell_service.rs
dmah42: i agree, 1 error per module.  however i add an asterisk "or service", so we should have a top-level error for Cell Service in this case.

i then think the utility classes and methods should return low-level errors with the service level translating them to the top-level error.  that's why i lean towards (1) in your list of options.

but you make a really good point about consuming Command and how that brings us a level of safety. i'm not sure how to build that into the clarity of error handling that i think the above policy would bring.

auraed/src/runtime/cells/cell.rs
future-highway: There is an argument to be made that `executables` can be its own module, then `cells` would depend on it, and `CellService` would depend on `cells`, but I think breaking them up would be overkill at this point. However, it does help think about why there are multiple error types.

Definitely interested in your thoughts on the below comment too (when you have a chance).

dmah42: but exactly: we wouldn't break them up into their own modules. which is why we shouldn't separate the errors either :)

auraed/src/runtime/cells/cell.rs
```diff
@@ -82,73 +80,83 @@ impl Cell {
...
+        let executable_name = executable.name.clone();
+
+        // Ignoring return value as we've already assured ourselves that the key does not exist.
+        let _ = self.executables.insert(executable_name.clone(), executable);
+
         // Start the child process
-        let exe =
-            Executable::start(exe_name.clone(), command, args, description)
-                .map_err(|e| CellError::ExecutableError {
+        if let Some(executable) = self.executables.get_mut(&executable_name) {
```
dmah42, line 103: we could not ignore the return value of insert and save a lookup here, no?

dmah42: ie, i don't know what we'd put in an "else" here so this feels a bit redundant.

auraed/src/runtime/cells/cell.rs
```diff
@@ -82,73 +80,83 @@ impl Cell {
...
-            });
-        }
+                }
+            })?;
 
-        info!(
-            "Cells: cell_name={} executable_name={exe_name} spawn() -> pid={}",
-            self.name, exe_pid.pid
-        );
+            // TODO: We've inserted the executable into our in-memory cache, and started it,
```
dmah42, line 114: maybe we should only insert after we've started and added it to the cell?

auraed/src/runtime/cells/cell.rs
```diff
@@ -172,14 +180,27 @@ pub(crate) enum CellError {
     ExecutableExists { cell_name: CellName, executable_name: ExecutableName },
     #[error("cell '{cell_name} could not find executable '{executable_name}'")]
     ExecutableNotFound { cell_name: CellName, executable_name: ExecutableName },
-    #[error("cell '{cell_name}': {source}")]
-    ExecutableError { cell_name: CellName, source: ExecutableError },
+    #[error("cell '{cell_name}' failed to start executable '{executable_name}' ({command:?}) due to: {source}")]
+    FailedToStartExecutable {
```
dmah42: should we take the Executable directly instead of all the fields?

dmah42 (dismissed):

auraed/src/runtime/cells/cell.rs
future-highway: Since we insert it into the `HashMap` on the right above, we've lost ownership of `executable`. Inside the `if` we have a `&mut` to executable. 

The lookup is technically fallible, but we know it won't fail because we just called `insert` and there is a `MutexGuard` on the entire `CellsTable`, so nothing can happen in between. 

The else would be `unreachable!("executable is guaranteed to be in the HashMap")`, which would panic if the assumptions of it definitely being there is wrong. I left that out, but it is probably safer to put that in.

auraed/src/runtime/cells/cell.rs
future-highway: We had it as start -> add (to cgroup) -> insert. All three are technically fallible, but insert is the one we have the most control over, and can know will succeed because we check that the `HashMap` doesn't `contain_key` right before.

I was thinking about what we would want the state of the system/cache be if something fails. By doing the insert first, if anything else fails, at least we still have a reference to the executable and can decide what to do with it.

Depending on our order of execution, some of our choices seem to be:
1. (Insert last) executable doesn't start, we end the request -> no big deal.
2. (Insert last) Executable started and in wrong Cell (`self.cgroup.add_task` fails), and cache doesn't have a reference to it -> how do we heal from this state?
3. (Insert last) Executable started and in correct Cell, but cache doesn't have a reference to it -> unlikely; insert is going to succeed
4.  (Insert first) we always have a reference no matter the outcome of the other calls -> if anything fails after, we have the most options
5. (Insert middle) executable starts, and we have a reference to it, but maybe in the wrong cell -> probably the same benefits of 4?

auraed/src/runtime/cells/cell.rs
future-highway: If we do option 5 (Insert middle) from the comment above, we could do that. As it is right now, we don't have ownership of `Executable` when this error occurs.

auraed/src/runtime/cells/cell.rs
future-highway: Added the `unreachable`

dmah42 (approved):

PR #175, opened 2022-12-18
Move logic from CellsService into Cells(Table) 

future-highway:
This change helps prevent out of order calls to `Cells` (renamed from `CellsTable`). 

In addition, it should reduce each `CellService` fn to only needing to take out a lock in `Cells` one time. That should be more efficient and prevent potential race conditions.

At this point, `CellService` has also become a thin proxy over the rest of the `cells` module.


auraed/src/runtime/cells/cells.rs
```diff
@@ -30,13 +30,16 @@
 
 use super::{Cell, CellName, CellsError, Result};
 use std::{collections::HashMap, sync::Arc};
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, MutexGuard};
 
-/// CellsTable is the in-memory store for the list of cells created with Aurae.
+// TODO (future-highway): would a RWLock be more performant?
```
dmah42, line 35: i don't think we have read-only access (yet?) so Mutex is likely fine.

dmah42: though if we have a way to dump out the cache from somewhere for logging (making this up) then we might want it. but even then i don't think Mutex here is going to be any sort of performance overhead.

auraed/src/runtime/cells/cells.rs
```diff
@@ -84,17 +92,63 @@ impl CellsTable {
...
+    where
+        F: FnOnce(&mut Cell) -> Result<R>,
+    {
         let mut cache = self.cache.lock().await;
+        get_mut(&mut cache, cell_name, f)
+    }
 
+    /// Remove and return the [Cell] keyed by [CellName] from [Cells].
+    /// Returns an error if the [CellName] does not exist in the cache.
+    pub async fn free(&self, cell_name: &CellName) -> Result<Cell> {
```
dmah42, line 105: it's a tiny bit odd to me that something called "free" returns a valid item.  i would expect that after "free" any Cell referenced is no longer valid to use (a hangover from C++ memory models i suppose).

perhaps this should be "remove" even though it's called by "free".

auraed/src/runtime/cells/cells.rs
```diff
@@ -45,35 +48,40 @@ pub(crate) struct CellsTable {
...
-        // TODO: replace with this when it becomes stable
-        // cache.try_insert(cell_name.clone(), cgroup)
-
-        // Check if there was already a cgroup in the table with this cell name as a key.
-        if cache.contains_key(&cell_name) {
-            return Err(CellsError::CellExists { cell_name });
-        }
-        // Ignoring return value as we've already assured ourselves that the key does not exist.
-        let _ = cache.insert(cell_name, cell.into());
+        let _ = allocate(&mut cache, cell_name, cell)?;
```
dmah42, line 60: on the offchance that a future change copies and mutates "cell" in `allocate` then we might want to return the Cell here from the function call instead of ignoring it.

auraed/src/runtime/cells/mod.rs
```diff
@@ -1,15 +1,16 @@
...
 use error::{CellsError, Result};
 use executable::Executable;
 use executable_name::ExecutableName;
 
 mod cell;
 mod cell_name;
 mod cell_service;
-mod cells_table;
+#[allow(clippy::module_inception)]
+mod cells;
```
dmah42, line 13: maybe cell_cache? :)

or cell_impl? (ew).

auraed/src/runtime/cells/cells.rs
future-highway: I agree. The more I think about it, the less I'm thinking we would benefit from RWLock; there should be a note in the next line, I just wasn't ready to remove the TODO.

auraed/src/runtime/cells/cells.rs
future-highway: We can't return `Cell` or a ref to it here because the lifetime of it is tied to the lock (line above). That was also why we needed `allocate_then` which takes a `Fn` that lets the caller have access to `&Cell`.

auraed/src/runtime/cells/mod.rs
future-highway: I think `Cells` is the best name for it (maybe `CellsCollection`? if we wanted to change it) and the clippy warning is just a result of keeping the `cells` module in `auraed` instead of its own crate, which I don't think in necessary yet.

I moved away from 'CellsTable' because to the user of `Cells`, its implementation is unimportant at this point. 'CellsTable' or 'CellsCache' feels like it's leaking the "unimportant" details.

auraed/src/runtime/cells/cells.rs
future-highway: Another option is to just return 'Result<()>` as nothing is using the result as of now.

future-highway: Changed it to return `Result<()>`

auraed/src/runtime/cells/mod.rs
future-highway: CellsController?

dmah42: it's fine as it is.

dmah42 (approved):

PR #179, opened 2022-12-18
Simplify `Cells` by moving lock to `CellService`

future-highway:
This should also make it a bit more straight forward for creating grpc `Response`s for cells.

### DIFF
--- a/.github/workflows/build-compile-test.yml
+++ b/.github/workflows/build-compile-test.yml
@@ -34,5 +34,5 @@ jobs:
         run: make build
       - name: Lint (make lint)
         run: make lint
-      - name: Tests (make test)
-        run: make test
+      - name: Tests (make testci)
+        run: make testci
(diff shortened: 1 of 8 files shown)


dmah42 (commented):
any chance the tests can come back?

future-highway:
> any chance the tests can come back?

maybe, I'll check

future-highway:
Ok, updated tests are back. Please see the TODO note in `impl Drop for Cells`. I used `Drop` to cleanup the groups that are created when the tests run.

future-highway:
Ok, the tests that can't be run in GH can now be marked with #[ignore]. `make test` will still run them, but `make testci` will not (neither will `cargo test`). We also have a `make prcheck` which is intended to mimic what GH checks, and that will run all tests (`make test`).

dmah42 (approved):

PR #183, opened 2022-12-20
Docs, Introduce namespace logic and prepare for /proc mount in new namespaces

krisnova:

### DIFF
--- a/auraed/src/runtime/cells/executable.rs
+++ b/auraed/src/runtime/cells/executable.rs
@@ -2,7 +2,6 @@ use crate::runtime::cells::validation::ValidatedCell;
 use crate::runtime::cells::ExecutableName;
 use cgroups_rs::CgroupPid;
 use std::io;
-use sys_mount::*;
 use tracing::info;
 use unshare::Command;
 use unshare::Error;
--- a/auraed/src/runtime/cells/executable.rs
+++ b/auraed/src/runtime/cells/executable.rs
@@ -1,8 +1,9 @@
 use crate::runtime::cells::validation::ValidatedCell;
 use crate::runtime::cells::ExecutableName;
 use cgroups_rs::CgroupPid;
-use tracing::info;
 use std::io;
+use sys_mount::*;
+use tracing::info;
 use unshare::Command;
 use unshare::Error;
 use unshare::ExitStatus;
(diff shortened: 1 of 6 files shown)


auraed/src/runtime/cells/executable.rs
```diff
@@ -84,14 +85,15 @@ impl Executable {
...
+
                 let command = command.unshare(&namespaces_to_unshare);
 
                 // Run 'pre_exec' hooks from the context of the soon-to-be launched child.
                 let command = {
                     let executable_name = self.name.clone();
                     unsafe {
                         command.pre_exec(move || {
-                            aurae_process_pre_exec(&executable_name)
+                            pre_exec(&executable_name, spec.clone())
```
dmah42, line 96: are we ignoring the potential error result here?

auraed/src/runtime/cells/executable.rs
```diff
@@ -135,11 +137,39 @@ impl Executable {
...
+        info!(
+            "CellService: pre_exec(): Supported filesystems: {:?}",
+            supported
+        );
+        let mount_result =
+            Mount::builder().fstype("proc").mount("proc", "/proc");
+        match mount_result {
+            Ok(mount) => {
+                // TODO We need to figure out how to manage this flag AFTER the command has been executed
+                // TODO Either on kill or stop we need to call this
```
dmah42, line 165: so we need a post_exec callback?

auraed/src/runtime/cells/executable.rs
future-highway: We are not. `command.pre_exec` takes a function that must return an `io::Result<()>` (i.e., `std::result::Result<(), io::Error>`), which `pre_exec` (the inner one) does. If `pre_exec` returns an error, it will bubble up when calling `command.spawn`, and a few lines down we have `let child = command.spawn()?;`

auraed/src/runtime/cells/executable.rs
krisnova: I think so -- The more I think about it the more I think we are going to be pre and post hooks as a service... unfortunately 

There is going to be a lot of logic that will end up in the pre and post hooks... a lot...

future-highway (approved):

PR #185, opened 2022-12-20
Make pki directory for certgen, if missing

future-highway:
As reported by rwxrick in the discord (and I ran into this myself just now as I needed to start from scratch), the `make pki` command requires a pki directory.


dmah42:
we can add pki to the .gitignore too I think

future-highway:
Good idea. Done. Add it looks like buf let us through this time

dmah42 (approved):

PR #189, opened 2022-12-22
WIP: Using nix

future-highway:
This is the first step on one path I was thinking (using std's Command and making calls using the nix crate).

I can't tell if anything is actually happening though (but there are no errors). Previously (when we had sys_mount's mount, if the drop wasn't used I would get to a point where I didn't know how to fix my vm (where Nóva's obs would have issues) and would need to reset my vm to a snapshot. Now, nothing seems to go wrong, even if I use sys_mount (which I checked that the source looks equivalent to the nix call anyway).

Edit:
A potential reason I am not seeing a change in the output is that `Command.spawn` uses [posix_spawn](https://man7.org/linux/man-pages/man3/posix_spawn.3.html), which in turn calls [fork](https://man7.org/linux/man-pages/man2/fork.2.html), which says "The child process is an exact duplicate of the parent process except for the following points:..."

https://github.com/rust-lang/rust/blob/2d8651a92761421b0437ffb44ba5670bea5ee1df/library/std/src/sys/unix/process/process_unix.rs#L56

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,7 +142,6 @@ dependencies = [
  "tracing-journald",
  "tracing-log",
  "tracing-subscriber",
- "unshare",
  "uuid",
  "validation",
  "validation_macros",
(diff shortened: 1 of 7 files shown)


future-highway:
see #192

PR #190, opened 2022-12-22
Use sh -c to run command

future-highway:
I set sh as "usr/bin/sh". Is it always that?


auraed/src/runtime/cells/executable.rs
```diff
@@ -41,8 +39,8 @@ impl Executable {
             ExecutableState::Started(child) => Ok((child.id() as u64).into()),
 
             ExecutableState::Init | ExecutableState::Stopped(_) => {
-                let mut command = Command::new(&self.command);
-                let command = command.args(&self.args);
+                let mut command = Command::new("/usr/bin/sh");
```
dmah42, line 42: https://unix.stackexchange.com/questions/416725/does-sh-have-to-be-in-the-bin-directory

POSIX means it must be in $PATH, nothing else.

dmah42 (approved):

auraed/src/runtime/cells/executable.rs
future-highway: If I change it to "/bin/sh", it still works on my system, but I'm going to assume that is just as unreliable.

Took your hint of POSIX and saw that there is a `command` command that functions like `which`, and prints out the full path with `command -v sh` ("/usr/bin/sh" in my case). 

It also has `command -p sh` where `-p` flag is documented as "use a default value for PATH that is guaranteed to find all of the standard utilities"

future-highway: I tested `Command::new("sh").args(&["-c", &self.command])` and it didn't work for me. This PR branches off the main branch that still uses the unshare crate's `Command`. Maybe that is the issue.

PR #193, opened 2022-12-24
Introduce skeleton pods service and api 🎉 

krisnova:


api/v0/runtime.proto
```diff
@@ -68,6 +65,31 @@ service Spawn {}
...
+/// will look at feel like one or more "containers".
+///
+/// Pods will run an OCI compliant container image.
+///
+/// A pod is a group of one or more containers with shared network and storage.
+service PodService {
+
+  rpc Allocate(AllocatePodRequest) returns (AllocatePodResponse) {}
+  rpc Start(StartContainerRequest) returns (StartContainerResponse) {}
+  rpc Stop(StopContainerRequest) returns (StopContainerResponse) {}
```
future-highway, line 78: I still prefer these to be more verbose in their naming. E.g., `AllocatePod`, `StartContainer`, `StopContainer`, `FreePod`.

Not a big deal though

future-highway (changes requested):
Blocking the merge because the macro is not generating the right deno code.

aka, I'm requesting changes from myself

auraescript/src/runtime.rs
```diff
@@ -0,0 +1,49 @@
+/* -------------------------------------------------------------------------- *\
```
dmah42, line 1: is this a generated file now?  it seems weird to have to repeat the contents of the proto.

dmah42: oh i guess we did that anyway.

dmah42 (dismissed):

future-highway (approved):
LGTM

PR #194, opened 2022-12-25
Nested auraed

future-highway:


auraed/src/runtime/cells/cell_service.rs
```diff
@@ -98,44 +102,100 @@ impl CellService {
     async fn start(
         &self,
         request: ValidatedStartExecutableRequest,
-    ) -> Result<StartExecutableResponse> {
-        let ValidatedStartExecutableRequest { cell_name, executable } = request;
+    ) -> std::result::Result<Response<StartExecutableResponse>, Status> {
```
dmah42, line 105: maybe next iteration: this should probably be on ExecutableService as it should be possible to start an executable without a cell with auraed.

auraed/src/runtime/cells/cell_service.rs
```diff
@@ -98,44 +102,100 @@ impl CellService {
...
+            let client_config = cells
+                .get(&child_cell_name, move |cell| cell.client_config())
+                .map_err(CellsServiceError::CellsError)?;
+
+            // TODO: Handle error
+            let client = AuraeClient::new(client_config)
+                .await
+                .expect("failed to create AuraeClient");
+
+            // TODO: This seems wrong.
```
dmah42, line 137: i think it's ok if you imagine this on a separate service for managing executables.

dmah42 (approved):

krisnova:
Whats the plan with `/crates`? Do we intend to keep the pattern? Not necessarily a blocker on the PR I just want to make sure I understand our long term direction so I can open up issues accordingly.

auraed/src/lib.rs
```diff
@@ -51,36 +51,30 @@
...
 )]
 #![warn(missing_debug_implementations,
-        // TODO: missing_docs,
-        trivial_casts,
-        trivial_numeric_casts,
-        unused_extern_crates,
-        unused_import_braces,
-        unused_results
-        )]
+// TODO: missing_docs,
```
dmah42, line 54: why did the indent change here?  not a big deal just curious if my autoformat is doing it wrong.

dmah42:
i have no concerns about the code (but there is a lot and it's a bunch to review at once so i may be missing something.. we can always fix anything later).

however, the code organization confuses me.  what does it mean to have a top-level `crates` directory alongside `auraed`?  is `auraed` the service implementation and `crates` the "internal library" functionality?  what controls the visibility and usage of the crates in `crates` so we don't end up with everything depending on everything else?

this flat structure concerns me from a maintainability of code point of view, essentially.  but i don't understand why some things are where they are (that's exactly my point tbh) enough to be able to provide some guidance on how else it could be organized.

auraed/src/runtime/cells/cell_service.rs
future-highway: Confirming my agreement from discord: I like the idea of an ExecutableService.

I'm not sure about an executable without a cell (that is, the top auraed, true pid 1, maybe shouldn't have that ability). I thought I read that systemd runs everything in a cgroup. I assume there is a reason? That is something I'm sure you and Nóva can answer.

auraed/src/lib.rs
future-highway: This has to be my or Nóva's automatic formatting (I assume ours are the same since we're both on clion). I edited it to put it back in a cleanup.

auraed/src/runtime/cells/cell_service.rs
dmah42: it does yes (https://opensource.com/article/20/10/cgroups) which allows "per service" resource allocation.  

i model the difference as the `ExecutableService` is the "just run this thing for me" service and the `CellService` is the "run this thing for me with this resource constraint" service.  perhaps we don't need to separate these and everything is run in a Cell but with optional constraints.  this just means we don't need to iterate on your work here :D

my conceptual framework is based on previous discussions around the API where we had "executable", "cell", and "pod" as different layers of "running a thing".  maybe the first is no longer necessary.

future-highway:
RE: `/crates`...

It's fairly common to see rust projects with a top-level `crates` directory. Every crate is in there, with the cargo.toml/lock too. So in our case it would include `aurae-proto`, `auraed`, and `auraescript` too. However, I wouldn't say it is idiomatic; many projects don't have one at all. I can't remember seeing a project have a split like we do (my doing).

Things I like about `/crates`...
1. When all crates are in there, it is easy to locate the source code, the top-level gets "random" files all the time.
2. Not really about `/crates`, but projects with it many time also have an `/examples` directory with crates showing how to use the project (in our case we have scripts though, which is nice and more applicable)

Things I don't like...
1. The project window in my IDE gets indented an additional level.

*I thought I'd have more likes and dislikes :)*

I'm going to collapse `aurae-cells` and `aurae-executables` back into `auraed` as I think it causes overhead for others at this point in time. The separation makes sense to me, and I really just needed it for myself while doing this PR (it helps me to have little isolation zones with few types that can be pulled in). I don't think having them separate has any significant advantage in build times, and I can only think of 1 potential technical reason to keep them separate (I'm wondering if integration tests will somehow not hang like the unit tests are, and rust makes it easy to have integration tests at the crate level).

`aurae-client` should stay standalone, as it is a close copy of what is happening in `auraescript`, and I want to refactor `auraescript` to use that too.

I think we should keep the split for now and put all "low-touch" crates in there (not in this PR). To me, I see those as:
* validation (won't change often)
* aurae-proto (fully generated)
* aurae-client (mostly generated, only changes when rpc calls change)
* auraescript (mostly generated, only changes when rpc calls change)

That would leave `auraed` as the only crate at the top, which is great because most of the work happens in there and in the api directory.

dmah42:
> RE: `/crates`...
> 
> It's fairly common to see rust projects with a top-level `crates` directory. Every crate is in there, with the cargo.toml/lock too. So in our case it would include `aurae-proto`, `auraed`, and `auraescript` too. However, I wouldn't say it is idiomatic; many projects don't have one at all. I can't remember seeing a project have a split like we do (my doing).
> 
> Things I like about `/crates`...
> 
> 1. When all crates are in there, it is easy to locate the source code, the top-level gets "random" files all the time.
> 2. Not really about `/crates`, but projects with it many time also have an `/examples` directory with crates showing how to use the project (in our case we have scripts though, which is nice and more applicable)
> 
> Things I don't like...
> 
> 1. The project window in my IDE gets indented an additional level.
> 
> _I thought I'd have more likes and dislikes :)_
> 
> I'm going to collapse `aurae-cells` and `aurae-executables` back into `auraed` as I think it causes overhead for others at this point in time. The separation makes sense to me, and I really just needed it for myself while doing this PR (it helps me to have little isolation zones with few types that can be pulled in). I don't think having them separate has any significant advantage in build times, and I can only think of 1 potential technical reason to keep them separate (I'm wondering if integration tests will somehow not hang like the unit tests are, and rust makes it easy to have integration tests at the crate level).
> 
> `aurae-client` should stay standalone, as it is a close copy of what is happening in `auraescript`, and I want to refactor `auraescript` to use that too.
> 
> I think we should keep the split for now and put all "low-touch" crates in there (not in this PR). To me, I see those as:
> 
> * validation (won't change often)
> * aurae-proto (fully generated)
> * aurae-client (mostly generated, only changes when rpc calls change)
> * auraescript (mostly generated, only changes when rpc calls change)


is it fair to say that /crates is for "common utility" crates then? ie, things that everything else in the project will reasonably depend on.

it's the "util" directory of the project. (all projects either grow large enough to have a "util" directory or die -- hamon's law :P )

> 
> That would leave `auraed` as the only crate at the top, which is great because most of the work happens in there and in the api directory.

i can imagine separating out each of the auraed subsystems into crates at some point, but i guess that can happen within the auraed directory.

auraed/src/runtime/cell_service/cells/cell.rs
```diff
@@ -0,0 +1,183 @@
...
+// In order to facilitate that immutability:
+// NEVER MAKE THE FIELDS PUB (OF ANY KIND)
+#[derive(Debug)]
+pub struct Cell {
+    name: CellName,
+    spec: CellSpec,
+    state: CellState,
+}
+
+// TODO: look into clippy warning
```
dmah42, line 47: tl;dr: it's due to an imbalance between the three variants where one is _much larger_ than the other, creating a potential issue with the memory layout.

i think it's ok to ignore this one as i don't see a better way of doing it (boxing the fields would be plausible but clunky).

dmah42 (approved):

krisnova (approved):

future-highway:
> is it fair to say that /crates is for "common utility" crates then? ie, things that everything else in the project will reasonably depend on.
> 
> it's the "util" directory of the project. (all projects either grow large enough to have a "util" directory or die -- hamon's law :P )
> 

I think that's a fair wait to think about it and a good mainframe to keep as we move forward.

> 
> i can imagine separating out each of the auraed subsystems into crates at some point, but i guess that can happen within the auraed directory.

Yeah, I do prefer the small/contained crates as I feel like they are little isolation zones that I can better keep in my head, but the different crates do add some mental overhead when working on multiple at a time. Especially after a build where the compiler might complain about a bunch of things, you have to pay attention to the paths the locate the file.

dmah42 (approved):

PR #201, opened 2022-12-28
fix blog links to ISO 8601

dmah42:
note that the `docs` changes are generated by `make docs` and maybe they shouldn't be checked into the repo.

the `mkdocs.yml` is the only change i made

### DIFF
--- a/docs/stdlib/index.md
+++ b/docs/stdlib/index.md
@@ -94,4 +94,4 @@ A notable exception to the public specification above is the Aurae projects pref
 The traditional convention that is meant to reduce the likelihood of future breaking changes and ease the creation of macros for generating code:
 
 - rpc methods (e.g., `StartWidget`) should have dedicated request and response messages named `StartWidgetResponse` and `StartWidgetResponse`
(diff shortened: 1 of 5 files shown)


docs/stdlib/index.md
```diff
@@ -94,4 +94,4 @@ A notable exception to the public specification above is the Aurae projects pref
 The traditional convention that is meant to reduce the likelihood of future breaking changes and ease the creation of macros for generating code:
 
 - rpc methods (e.g., `StartWidget`) should have dedicated request and response messages named `StartWidgetResponse` and `StartWidgetResponse`
-- objects (e.g., `Widget`) should be embedded directly into their corresponding `StartWidgetRequest`, `StopWidgetRequest`, etc style methods.
+- objects (e.g., `Widget`) should be embedded directly into their corresponding `StartWidgetRequest`, `StopWidgetReqyest`, etc style methods.
```
alexschrod, line 97: This looks like a mistake to me; `Request` seems more correct than `Reqyest`.

dmah42: agreed.. related to the above where the typo fix in main was made in the generated file but not the source.

krisnova (approved):

PR #204, opened 2022-12-28
Cells cache reconciliation

### LINKED ISSUE #198: Possible to destroy cgroups manually and Aurae still has them in cache
Right now it is possible to use the `rmdir` command in the `/sys/fs/cgroup` directory to manually destroy cgroups. 

In the event that we manually destroy a cgroup that was started with `auraed` the internal cache still believes the cgroup exists. This is problematic when we try to re-create the cgroup again with `auraed`.

future-highway:
Introduced 2 more errors to CellsError:
1. `CgroupNotFound` -> We cached a `cell`, but the `cgroup` is not on the fs
2. `CgroupIsNotACell` -> A `cgroup` exists on the fs, but we don't have a cached cell

Anytime `CgroupNotFound` is encountered, we remove the cell from the cache, so repeated calls won't see the error multiple times.

close #198

### DIFF
--- a/auraed/src/runtime/cell_service/cells/cells.rs
+++ b/auraed/src/runtime/cell_service/cells/cells.rs
@@ -28,12 +28,12 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
-use super::{Cell, CellName, CellSpec, CellsError, Result};
+use super::{cgroups::Cgroup, Cell, CellName, CellSpec, CellsError, Result};
 use std::collections::HashMap;
 
 type Cache = HashMap<CellName, Cell>;
 
-/// Cells is the in-memory store for the list of cells created with Aurae.
+/// The in-memory cache of cells ([Cell]) created with Aurae.
 #[derive(Debug, Default)]
 pub struct Cells {
     cache: Cache,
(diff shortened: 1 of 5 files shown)


auraed/src/runtime/cell_service/cells/cells.rs
```diff
@@ -46,24 +46,30 @@ pub struct Cells {
...
+            } else {
+                Err(CellsError::CgroupIsNotACell {
+                    cell_name: cell_name.clone(),
+                })
+            };
         }
 
-        // `or_insert` will always insert as we've already assured ourselves that the key does not exist.
+        // From here, we know the cgroup doesn't exist, so remove from cache if it does
+        let _ = self.cache.remove(&cell_name);
```
dmah42, line 71: should we warn or something? aiui this is user error: they removed the cgroup from the FS without calling `free` on auraed.

auraed/src/runtime/cell_service/cells/cells.rs
```diff
@@ -116,6 +131,25 @@ impl Cells {
...
+    fn handle_cgroup_does_not_exist(
+        &mut self,
+        cell_name: &CellName,
+    ) -> Result<()> {
+        if Cgroup::exists(cell_name) {
+            return Ok(());
+        }
+
+        let Some(_removed) = self.cache.remove(cell_name) else {
+            // Cell doesn't exist & cgroup doesn't exit
```
dmah42, line 144: nit: s/exit/exist

auraed/src/runtime/cell_service/cells/error.rs
```diff
@@ -54,4 +54,8 @@ pub enum CellsError {
     FailedToKillCellChildren { cell_name: CellName, source: io::Error },
     #[error("cell '{cell_name}' could not be freed: {source}")]
     FailedToFreeCell { cell_name: CellName, source: cgroups_rs::error::Error },
+    #[error("a cgroup '{cell_name}' exits on the system, but is not controlled by auraed")]
+    CgroupIsNotACell { cell_name: CellName },
+    #[error("cell '{cell_name}' did not find cgroup on system")]
```
dmah42, line 59: alternative "cgroup '{cell_name}' not found on host"

auraed/src/runtime/cell_service/error.rs
```diff
@@ -54,7 +54,9 @@ impl From<CellsServiceError> for Status {
                 CellsError::FailedToAllocateCell { .. }
                 | CellsError::AbortedAllocateCell { .. }
                 | CellsError::FailedToKillCellChildren { .. }
-                | CellsError::FailedToFreeCell { .. } => Status::internal(msg),
+                | CellsError::FailedToFreeCell { .. }
+                | CellsError::CgroupIsNotACell { .. }
```
dmah42, line 58: are these internal errors? CgroupNotFound should probably be a `not_found` error. CgroupIsNotACell is probably a `bad_request`.

future-highway: I don't see a `bad_request`. I think `failed_precondition` might be the one (that's what we use for validation errors).

auraed/src/runtime/cell_service/cells/cells.rs
future-highway: Barring logic errors in the code, I think this has to be user error as well. The user must have used aurae to create the cell and then somehow the cgroup is no longer on the host.

I don't know the correct response, but I put a `warn!` for now and a TODO (happy to change the logic in this PR if we want to make a decision)

> // TODO: Should we not remove the cell (that has no cgroup) from the cache and
    //       force the user to call Free? Free will also return an error, but we may be
    //       calling other logic in free that we want to run.

I also made cell's drop call `free` (not just in tests).

auraed/src/runtime/cell_service/cells/cells.rs
```diff
@@ -28,42 +28,54 @@
...
+                })
+            };
+        }
+
+        // From here, we know the cgroup doesn't exist, so remove from cache if it does
+        if let Some(_removed) = self.cache.remove(&cell_name) {
+            // TODO: Should we not remove the cell (that has no cgroup) from the cache and
+            //       force the user to call Free? Free will also return an error, but we may be
+            //       calling other logic in free that we want to run.
+            warn!("found cached cell ('{cell_name}') without cgroup");
```
dmah42, line 76: maybe add "did you forget to call free on the cell?" or something

dmah42 (dismissed):

dmah42 (approved):

PR #206, opened 2022-12-29
capture and log stdout and stderr from executables

### LINKED ISSUE #173: Plumb stdout and stderr through CellService
I believe we will need to develop a logging cache that serves as an in-memory caching layer (Note: in the future it will need to also persist to disk!) for both stdout and stderr streams from an executable within a cell.

We will need to be able to hook into a stream at any moment and have some basic guarantees about the data and how we retrieve it.

dmah42:
Fixes #173 

In the case of a spawned `Child`, this reads stdout directly.  In the case of a cloned `Process`, this gets the stdout or stderr file descriptor from the process and reads from it.  `Process` is responsible for reading the stdout or stderr.

The `Executable` is responsible for reading the output is read in batches of lines in a thread.

The output is then sent for both tracing output in a "running process" span and to new `LogChannel`s that are created when an `Executable` is created.  this latter will allow the `ObserveService` to expose the logs to the user, though this hasn't been implemented.  The span includes the name of the executable that's running.

The output is always logged at `INFO` level, but with a `level` (and `channel`) field indicating whether it came from stdout or stderr.

### DIFF
--- a/auraed/src/logging/log_channel.rs
+++ b/auraed/src/logging/log_channel.rs
@@ -47,11 +47,17 @@ pub struct LogChannel {
 
 impl LogChannel {
     /// Constructor creating the channel for log communication
-    pub fn new(name: &str) -> LogChannel {
+    pub fn new(name: String) -> LogChannel {
         // TODO: decide for a cap. 40 is arbitrary
         let (producer, _) = broadcast::channel(40);
-        LogChannel { producer, name: name.to_string() }
+        LogChannel { producer, name }
     }
+
+    /// Getter for the name
+    pub fn name(&self) -> &String {
+        &self.name
+    }
+
     /// Getter for producer channel
     pub fn get_producer(&self) -> Sender<LogItem> {
         self.producer.clone()
@@ -63,11 +69,11 @@ impl LogChannel {
     }
 
     /// Wrapper that sends a log line to the channel
-    pub fn log_line(producer: Sender<LogItem>, line: &str) {
+    pub fn log_line(producer: Sender<LogItem>, line: String) {
         // send returns an Err if there are no receivers. We ignore that.
         let _ = producer.send(LogItem {
             channel: "unknown".to_string(),
-            line: line.to_string(),
+            line,
             // TODO: milliseconds type in protobuf requires 128bit type
             timestamp: get_timestamp_sec(),
         });
@@ -105,24 +111,24 @@ mod tests {
     #[tokio::test]
     async fn test_ringbuffer_queue() {
         init_logging();
-        let lrb = LogChannel::new("Test");
+        let lrb = LogChannel::new("Test".into());
         let prod = lrb.get_producer();
         let mut consumer = lrb.get_consumer();
 
-        LogChannel::log_line(prod.clone(), "hello");
-        LogChannel::log_line(prod.clone(), "aurae");
-        LogChannel::log_line(prod.clone(), "bye");
+        LogChannel::log_line(prod.clone(), "hello".into());
+        LogChannel::log_line(prod.clone(), "aurae".into());
+        LogChannel::log_line(prod.clone(), "bye".into());
 
         let cur_item = LogChannel::consume_line(&mut consumer).await;
         assert!(cur_item.is_some());
-        assert_eq!(cur_item.unwrap().line, "hello");
+        assert_eq!(cur_item.unwrap().line, "hello".to_string());
 
         let cur_item = LogChannel::consume_line(&mut consumer).await;
         assert!(cur_item.is_some());
-        assert_eq!(cur_item.unwrap().line, "aurae");
+        assert_eq!(cur_item.unwrap().line, "aurae".to_string());
 
         let cur_item = LogChannel::consume_line(&mut consumer).await;
         assert!(cur_item.is_some());
-        assert_eq!(cur_item.unwrap().line, "bye");
+        assert_eq!(cur_item.unwrap().line, "bye".to_string());
     }
 }
--- a/auraed/src/logging/log_channel.rs
+++ b/auraed/src/logging/log_channel.rs
@@ -54,7 +54,7 @@ impl LogChannel {
     }
 
     /// Getter for the name
-    pub fn name(&self) -> &String {
+    pub fn name(&self) -> &str {
         &self.name
     }
 
--- a/auraed/src/runtime/cell_service/executables/executable.rs
+++ b/auraed/src/runtime/cell_service/executables/executable.rs
@@ -98,13 +98,12 @@ impl Executable {
     /// Stops the executable and returns the [ExitStatus].
     /// If the executable has never been started, returns [None].
     pub fn kill(&mut self) -> io::Result<Option<ExitStatus>> {
-        let exit_status: Option<ExitStatus>;
-        {
+        let exit_status = {
             let mut inner = self.inner.lock().map_err(|e| {
                 io::Error::new(io::ErrorKind::Other, e.to_string())
             })?;
             match &mut inner.state {
-                ExecutableState::Init { .. } => exit_status = None,
+                ExecutableState::Init { .. } => None,
                 ExecutableState::Started { process, .. } => {
                     let proc_status: ExitStatus;
                     {
@@ -115,11 +114,11 @@ impl Executable {
                         proc_status = proc.wait()?;
                     }
                     inner.state = ExecutableState::Stopped(proc_status);
-                    exit_status = Some(proc_status);
+                    Some(proc_status)
                 }
-                ExecutableState::Stopped(status) => exit_status = Some(*status),
-            };
-        }
+                ExecutableState::Stopped(status) => Some(*status),
+            }
+        };
         self.log_thread
             .take()
             .map(thread::JoinHandle::join)
@@ -150,11 +149,10 @@ impl Executable {
 fn spawn_log_thread(
     inner: Arc<Mutex<ExecutableInner>>,
 ) -> thread::JoinHandle<io::Result<()>> {
-    let local_inner = inner;
     thread::spawn(move || -> io::Result<()> {
         let mut running = true;
         while running {
-            let inner = local_inner.lock().map_err(|e| {
+            let inner = inner.lock().map_err(|e| {
                 io::Error::new(io::ErrorKind::Other, e.to_string())
             })?;
             let _span =
--- a/auraed/src/runtime/cell_service/executables/executable.rs
+++ b/auraed/src/runtime/cell_service/executables/executable.rs
@@ -1,16 +1,27 @@
 use super::process::Process;
 use super::{ExecutableName, ExecutableSpec};
+use crate::logging::log_channel::LogChannel;
 use nix::sys::signal::SIGKILL;
 use nix::unistd::Pid;
 use std::ffi::OsString;
-use std::process::Command;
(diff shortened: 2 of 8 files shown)


auraed/src/logging/log_channel.rs
```diff
@@ -47,11 +47,17 @@ pub struct LogChannel {
...
-    pub fn new(name: &str) -> LogChannel {
+    pub fn new(name: String) -> LogChannel {
         // TODO: decide for a cap. 40 is arbitrary
         let (producer, _) = broadcast::channel(40);
-        LogChannel { producer, name: name.to_string() }
+        LogChannel { producer, name }
     }
+
+    /// Getter for the name
+    pub fn name(&self) -> &String {
```
future-highway, line 57: This should be &str. I don't think &String is ever really used. 

*pulling down the pr to review the rest*

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -54,12 +76,41 @@ impl Executable {
...
+                    );
+                    LogChannel::log_line(
+                        self.stdout.get_producer().clone(),
+                        line.to_string(),
+                    );
+                }
+                for line in self.read_stderr()? {
+                    event!(
+                        Level::INFO,
+                        level = "error",
```
future-highway, line 105: I think you meant to have `Level::INFO` and `level = "error"` but commenting just in case.

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -75,6 +126,117 @@ impl Executable {
 
         Some(process.pid())
     }
+
+    /// Returns any unread lines from stdout if [Executable] is running, otherwise returns an empty
+    /// [Vec].
+    pub fn read_stdout(&mut self) -> io::Result<Vec<String>> {
```
future-highway, line 132: I think these should be on `Process`. Access to the inner `procfs::process::Process` or `Child` is a result of `Process` being an enum and enum variants always(?) being pub.

As of now, `Executable` always results a `Process::Spawned` and `NestedAuraed` in a `Process::Cloned`. Adding to `Process` allows us to share logic between both (e.g., `kill`).

We should probably make `Process` in a struct with an inner emum (`ProcessInner` or something) to guide the correct usage.

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -31,7 +42,16 @@ enum ExecutableState {
 impl Executable {
     pub fn new<T: Into<ExecutableSpec>>(spec: T) -> Self {
         let ExecutableSpec { name, description, command } = spec.into();
-        Self { name, description, state: ExecutableState::Init { command } }
+        let stdout = Arc::new(LogChannel::new(format!("{}::stdout", name)));
+        let stderr = Arc::new(LogChannel::new(format!("{}::stderr", name)));
```
future-highway, line 46: I'm wondering why these are `Arc` wrapped. I haven't tried changing anything to see, but I feel like they shouldn't be.

LogChannel in general has a weird interface that we should fix.

dmah42: so i can pass them into the thread (which is currently commented out).

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: i went back and forth.. it felt weird to have errors in the server logs caused by user script output, so i settled on info level logging with an inner "level" to indicate the inner logging level.  i could also just have "channel" or "stream".

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: moving them.. this might also help solve my thread issue as the process can be Arc i think and we can access it from a thread running on Executable's scope.

auraed/src/runtime/cell_service/executables/executable.rs
future-highway: LogChannel should be clone-able (if we add `Clone` to the derive). `Sender<LogItem>` is expected to be cloned and I assume `name` will always be small enough to be cheap to clone.

We can refactor later though.

dmah42: yeah .. i might need an ExecutableInner anyway to allow the Process to be shared with the logging thread too. so i think we'll need 

```
struct Executable {
  inner: Arc<Mutex<ExecutableInner>>;
};

struct Executable {
  process: Process;
  stdout: LogChannel;
 ...
};
```

at which point everything gets a bit more complex but also easier.

future-highway: Do we not want the same functionality on all `Process` wrappers (i.e., `Executable` & `NestedAurae`)?  In which case, I think, we would continue to expand the functionality in `Process`, and the wrappers would proxy to it, like with `kill` and now `read_stdout/err`

dmah42: i thought Process should be the low level view.. ie, exposing information but not doing anything with it. then Executable (and presumably NestedAurae) can decide what to do with that data.  so Executable will push it to Observe (and optionally trace) whereas NestedAurae would only trace it.

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: we now have inner level and channel and a span to hold them, which i think means "info" is ok.

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: i went with the `ExecutableInner` and it's much cleaner.

auraed/src/runtime/cell_service/executables/executables.rs
```diff
@@ -90,9 +91,9 @@ impl Executables {
 
         let Some(exit_status) = exit_status else {
             // Exes that never started return None
-            let executable = self.cache.remove(executable_name).expect("exe in cache");
+            let _ = self.cache.remove(executable_name).expect("exe in cache");
             return Err(ExecutablesError::ExecutableNotFound {
-                executable_name: executable.name,
+                executable_name: executable_name.clone(),
```
future-highway, line 96: Why introduce a clone here? We know that the exe is in the cache (hence the use of `expect("exe in cache")`) since we didn't error in `self.get_mut` which is the first call in `stop`.

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -23,66 +38,179 @@ enum ExecutableState {
...
-        Some(process.pid())
+        Ok(Some(proc.pid()))
     }
 }
 
+/// Spawns a thread that produces log lines while the [Executable] is running.
+fn spawn_log_thread(
+    inner: Arc<Mutex<ExecutableInner>>,
+) -> thread::JoinHandle<io::Result<()>> {
+    let local_inner = inner;
```
future-highway, line 153: This doesn't seem needed. `inner` should move into the closure without issue. No?

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -23,66 +38,179 @@ enum ExecutableState {
...
+        while running {
+            let inner = local_inner.lock().map_err(|e| {
+                io::Error::new(io::ErrorKind::Other, e.to_string())
+            })?;
+            let _span =
+                info_span!("running process", name = ?inner.name).entered();
+            match &inner.state {
+                ExecutableState::Init { .. } => {}
+                ExecutableState::Started { process, .. } => {
+                    let mut proc = process.lock().map_err(|e| {
```
future-highway, line 165: Won't this block `kill()` and `pid()` until there is nothing left to read from stdout/err? Is that ok?

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -23,66 +38,179 @@ enum ExecutableState {
...
-            ExecutableState::Init { .. } => Ok(None),
-            ExecutableState::Started { process, .. } => {
-                process.kill(Some(SIGKILL))?;
-                let exit_status = process.wait()?;
-                self.state = ExecutableState::Stopped(exit_status);
-                Ok(Some(exit_status))
-            }
-            ExecutableState::Stopped(exit_status) => Ok(Some(*exit_status)),
+    pub fn kill(&mut self) -> io::Result<Option<ExitStatus>> {
+        let exit_status: Option<ExitStatus>;
```
future-highway, line 101: Nit: this block is written in an unexpected way. The following should be equivalent and safer against programmer error:

```rust
let exit_status = {
            let mut inner = self.inner.lock().map_err(|e| {
                io::Error::new(io::ErrorKind::Other, e.to_string())
            })?;
            match &mut inner.state {
                ExecutableState::Init { .. } => None,
                ExecutableState::Started { process, .. } => {
                    let proc_status = {
                        let mut proc = process.lock().map_err(|e| {
                            io::Error::new(io::ErrorKind::Other, e.to_string())
                        })?;
                        proc.kill(Some(SIGKILL))?;
                        proc.wait()?
                    };
                    inner.state = ExecutableState::Stopped(proc_status);
                    Some(proc_status)
                }
                ExecutableState::Stopped(status) => Some(*status),
            }
        };
```

future-highway (approved):
I think it's fine to merge. I'm not sure if the cause is related to this PR, but I did notice a delay in stopping auraed in the terminal.

Generally, I think we need to refactor LogChannel and try switching the tokio and making these calls async.

auraed/src/runtime/cell_service/executables/executables.rs
dmah42: otherwise it would be a move from a shared reference.  and i can't get the name from the executable anymore without a lock, which i'd like to avoid in this flow.

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: yeah, this was a holdover.

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: no. we don't read until "end", we just read whatever's currently in the pipe before unlocking.

it's possible that if the running executable is spamming stuff we might see a lock being held for longer.  maybe it's worth changing the process read to read a fixed amount (1k?) into a buffer, though then we might split log lines...

dmah42:
my rebase ended up creating the Pod stuff in the generated code.. should we be not checking in the gen code?

dmah42:
> I think it's fine to merge. I'm not sure if the cause is related to this PR, but I did notice a delay in stopping auraed in the terminal.
> 

i think this is the fault of this PR and i'm not sure exactly where it is. i think because the join needs to wait for the log thread to finish up (10ms + whatever read happens) it adds a delay.

> Generally, I think we need to refactor LogChannel and try switching to tokio and making these calls async.

that would be great. i think it's just a refactor of the log thread into a tokio spawn with the command being a tokio command, right?

future-highway (approved):
I don't think we should worry about the delay at this stage. We can see if a solution emerges post a refactor. I think you are on point on with what needs to be done in the refactor (perhaps a little more).

---

RE gen code: Yes, we have been checking it in, though I'm not sure it has made a difference (buf is being run anyway). I think it is just too easy to miss checking them in.

Issue #79, opened 2022-10-24
Create and use build container (speed up the builds)


krisnova:
Right now we have several build jobs that are executed during a pull request and a merge to main

 - [build-docs.yml](https://github.com/aurae-runtime/aurae/blob/main/.github/workflows/build-docs.yml)
 - [build-compile-test.yml](https://github.com/aurae-runtime/aurae/blob/main/.github/workflows/build-compile-test.yml)
 - [deploy-website.yml](https://github.com/aurae-runtime/aurae/blob/main/.github/workflows/deploy-website.yml)

Can we please leverage #78 and have the builds use a container instead of `apt-get` installing the dependencies each iteration?

sidpalas:
I took a stab at this with https://github.com/sidpalas/aurae/pull/1

It still needs some work and the container-based version is slower than the one without. Does anyone have thoughts on why this is? Are containers run with constrained CPU/memory relative to the default VM?

future-highway:
The build step in the two jobs are doing different work (container one is downloading crates). That would account for the some difference in run times. 

They both were cache hits (so not cold builds) but the container based one either didn't have everything it needed cached, or it isn't configured (or can't) take advantage of the cache as well as the other. 

The biggest take away for me though was that initializing the container was more time then downloading the dependencies (the current setup). So it may be worth waiting to switch until containers can provide some other value or the dependencies increase, flipping the advantage.

sidpalas:
@future-highway, that makes sense! 

I added another test job using the busybox image (1.24MB) and it starts in [2 seconds](https://github.com/sidpalas/aurae/actions/runs/3378985643/jobs/5609941728) to see how long initialization would take for a small image, so it seems that most of the container initialization time is spent pulling the large image with the Rust toolchain.

With self-hosted runners, this image could be cached across runs, but with GitHub's hosted runners I don't think we have any control over image caching.

Holding off until there is a significant speedup seems like the way to go.

Also, I am not very familiar with Rust so if others have ideas on making the build container image smaller

taliamax:
> @future-highway, that makes sense!
> 
> I added another test job using the busybox image (1.24MB) and it starts in [2 seconds](https://github.com/sidpalas/aurae/actions/runs/3378985643/jobs/5609941728) to see how long initialization would take for a small image, so it seems that most of the container initialization time is spent pulling the large image with the Rust toolchain.
> 
[...]

Issue #78, opened 2022-10-24
Aurae needs a container registry


krisnova:
The projects needs a place to store containers, which could implicitly mean the project has an authz/secrets problem as well. 

Can we please identify a place for the project to store containers? Can we also work with the other maintainers to make sure folks and the build systems have access to the container registry as needed?

jesse-peters:
I've been using the GHA package repository for the build image, and I think it might be a good starting place. Auth is relatively simple since its using GHA to push, and its public for anyone who wants to pull the images.

PR #209, opened 2022-12-29
refactor: Add build GHA image

jesse-peters:
I took another whack at [this](https://github.com/aurae-runtime/aurae/issues/79) using a few different methods. The time to run is still a bit longer than native GHA jobs, but with a self-hosted runner I think it would be pretty similar.

I'm building the image as the first step in the pipeline, then using that image to run the jobs on downstream. I have introduced [cargo-chef](https://crates.io/crates/cargo-chef) to package the dependencies, without needing to compile the code itself.

Using minideb we launch a container that has the rust requirements (but without extras like cargo-chef). This runs the actual make steps, and uses native GHA caching for the rust modules.

I would be happy to setup and own a self-hosted runner and [docker registry](https://github.com/aurae-runtime/aurae/issues/78). I'm happy to chat about that in discord too (kuxaku) :)

Most of the build layers should cache across branches too (using buildkits native GHA cache), so the first time it runs should be the longest.

Biggest Risks:
- I got the caching wrong and the pipeline doesn't have correct dependencies someplace along the test/build process
- The build times are longer, impacting developer feedback times

Out of scope:
- I did not apply the docker image to the `deploy` step yet.

I kept the original two GHA jobs in the pipeline for reference. If/when this gets merged they can be ran alongside for a bit to ensure its working, or deleted in a second PR.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.cache
+.dockerignore
+.github*
+.gitignore
+.idea*
+Dockerfile
+docs/crate/*
+target*
--- a/.github/workflows/build-with-containers.yml
+++ b/.github/workflows/build-with-containers.yml
@@ -45,7 +45,7 @@ jobs:
         uses: docker/login-action@v2
         with:
           registry: ghcr.io
-          username:  ${{ github.repository_owner }}
+          username:  ${{ github.actor }}
           password:  ${{ secrets.GITHUB_TOKEN }}
 
       - name: Create Docker Build Image
@@ -70,7 +70,7 @@ jobs:
     container:
       image: ghcr.io/${{ github.repository }}/aurae-builder:${{ needs.build-container.outputs.tag }}
       credentials:
-        username:  ${{ github.repository_owner }}
+        username:  ${{ github.actor }}
         password:  ${{ secrets.GITHUB_TOKEN }}
       options: --cpus 1
     steps:
@@ -101,7 +101,7 @@ jobs:
     container:
       image: ghcr.io/${{ github.repository }}/aurae-builder:${{ needs.build-container.outputs.tag }}
       credentials:
-        username:  ${{ github.repository_owner }}
+        username:  ${{ github.actor }}
         password:  ${{ secrets.GITHUB_TOKEN }}
       options: --cpus 1
     steps:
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,37 @@
+# syntax = docker/dockerfile:1.4
```
krisnova, line 1: Do we want this to be the "root" Dockerfile for the project? Would it make sense to have a `/docker` or `/images` or `/containers` directory where we can house multiple Dockerfiles for different reasons? What about plumbing the docker commands through to the Makefile and having an `$OCI` variable so we can build with out container tools as well?

Something like this:

```make 
OCI=docker
$(OCI) build -t $(NAME) .
```

jesse-peters: You beat me to it - I was mid writing up a comment and was going to ask if this was the desired location. I'll move it and add the docker commands via the make file too.

jesse-peters: I'm not using any native docker commands in the pipeline, just pre-defined GHA jobs. I have added the ability to build the docker container to the Makefile.

jesse-peters:
Sorry for the noise, I created this PR a little bit early it looks like

krisnova:
Want to squash for us?

jesse-peters:
Happy to. Bare with me, git and I dont get alone sometimes :) (I usually have squash turns on automatically in my repos)

future-highway:
@cla-bot check

jesse-peters:
I think I'll need a project maintainer to add write permissions containers to this project. I'm using `aurae-builder` as the name of the package. Now is a good time to change that too if we wanted to use a different one.

I think its here for an org:

https://github.com/orgs/aurae-runtime/packages

https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-a-registry-using-a-personal-access-token

jesse-peters:
@cla-bot check

krisnova:
Hey @jesse-peters I hate to sound like the idiot who doesn't understand how to use GitHub, but I have spent the last 30 minutes trying to understand what it is I am supposed to do to help you out here and I honestly am stumped.

What exactly do you need? Can you be as descriptive as possible? Is there any way you can share a "this specific thing is failing and if you make this specific thing go away we will be happy" style description?

PR #213, opened 2022-12-29
Docs improvements

megamorf:
This PR does two things:

1. Fixes the "Edit this page" link that mkdocs provides for convenience:

image

2. Adds vale (a _syntax-aware linter for prose_) as tool  to ensure a certain level of consistency in the docs.

Let me know if you're fine with the addition - as it is now it would cause the `Documentation` workflow to fail if a change in the docs violates the style that vale enforces.


megamorf:
The CI failed due a rate limiting issue.

krisnova (approved):

Issue #199, opened 2022-12-27
POSIX Signal Handler and Proxy/Bus


krisnova:
Now that `auraed` is launching nested versions of itself, we will need to proxy POSIX signals through `auraed` 

1. Implement a POSIX compliant signal handler in Rust in Auraed and provide documentation on which signals map to what behavior.

`SIGKILL` should terminate (kill) the process
`SIGHUP` should reload the config from disk and reopen logfiles 
`SIGINT` should "interrupt" the process and begin to "die nice" ensuring any cleanup logic can be done 

2. Use `SIGINT` instead of `SIGKILL` to "free" a nested auraed after the signal handler has been implemented.

3. Proxy all [signals](https://dsa.cs.tsinghua.edu.cn/oj/static/unix_signal.html) to nested executables for them to manage independently. 

For example sending a `SIGHUP` to a nested `auraed` should proxy a `SIGHUP` to all of the nested executables!

dmah42:
i believe `SIGTERM` should do the "die nice".  `SIGINT` is a "user triggered interrupt" signal that processes can interpret however they want but it doesn't necessarily suggest termination.

https://dsa.cs.tsinghua.edu.cn/oj/static/unix_signal.html

this also suggests "SIGINT is nearly identical to SIGTERM".  `SIGINT` is explicity "sent from a user" though.

dmah42:
if a SIGKILL is sent to auraed, i assume it means "die in a fire" and we should take down everything auraed has spawned along with us.  

SIGTERM would be the same, and would also expect spawned processes/nested aurae to "die nicely" so that needs to be passed along, but maybe that means calling the various `Stop` and `Free` APIs instead.

SIGHUP is a different one as i suspect we'd _only_ want that to be handled by auraed.

so i think 1 and 2 hold, but 3 _isn't_ necessary as it's not obvious how we would interpret the signals.

this leaves an open question: how do we send a signal to a spawned executable managed by a nested auraed?

PR #214, opened 2022-12-30
Fix call to `wait` done post `SIGKILL`

future-highway:


dmah42:
i think SIGINT is not appropriate as it is a "user interrupt request".  i would expect a parent to send SIGTERM to a child process as it gives an opportunity to clean up (unlike SIGKILL).  SIGQUIT is right out, given the core dump it generates.

have we considered SIGTERM?

future-highway:
I think we should use whatever is most expected (i.e., standard practice). #199 seemed to indicate we want to use `SIGINT` but https://aurae.io/signals/ says to use `SIGTERM`.

dmah42:
> I think we should use whatever is most expected (i.e., standard practice). #199 seemed to indicate we want to use `SIGINT` but https://aurae.io/signals/ says to use `SIGTERM`.

https://dsa.cs.tsinghua.edu.cn/oj/static/unix_signal.html <-- SIGTERM :)

auraed/src/runtime/cell_service/executables/auraed/nested_auraed.rs
```diff
@@ -215,11 +215,14 @@ impl NestedAuraed {
         //       (contrary to ^^^). What do we want to signal?
 
         if wants_isolated_pid(&self.shared_namespaces) {
-            // TODO: Here, SIGINT works when using auraescript, but hangs(?) during unit tests.
-            //       SIGKILL, however, works in both scenarios. Why?
-            self.process.kill(Some(SIGKILL))?;
+            // TODO: Here, SIGTERM works when using auraescript, but hangs(?) during unit tests.
```
dmah42, line 218: sigterm hangs too? hm.

auraed/src/runtime/cell_service/executables/auraed/nested_auraed.rs
```diff
@@ -215,11 +215,14 @@ impl NestedAuraed {
...
+            //       SIGKILL, however, works in both scenarios. This may be solved once we proxy
+            //       signals to nested auraeds.
+            self.process.kill(Some(SIGTERM))?;
             self.process.wait()
         } else {
+            // TODO: https://aurae.io/signals/ seems to indicate that we always want
+            //       to send SIGTERM to nested auraeds?
             self.process.kill(SIGKILL)?;
```
dmah42, line 226: i think SIGTERM too, but we should also handle that signal and shut down as cleanly as possible.

dmah42 (dismissed):

dmah42 (approved):

PR #223, opened 2022-12-31
Proxy SIGTERM to nested auraeds

future-highway:
Added a SIGTERM signal handler that will broadcast to all cells.

Notes:
* The `Cells` cache is no longer on `CellService` so that we can access it anywhere in auraed (including from the signal handlers).
* `Drop` on `Cell` will result in a `SIGKILL` to `nested_auraed` when the `Cell` is still in the `Allocated` state. I don't expect this to be called (except in tests), but it gives us some more assurance that resources will be cleaned up.

Future work (in a different PR) will be to make this all async.

RE: #199

### DIFF
--- a/auraed/src/runtime/cell_service/cells/mod.rs
+++ b/auraed/src/runtime/cell_service/cells/mod.rs
@@ -35,9 +35,6 @@ pub use cells::Cells;
 use cgroups::Cgroup;
 pub use cgroups::{CgroupSpec, CpuCpus, CpuQuota, CpuWeight, CpusetMems};
 pub use error::{CellsError, Result};
-use lazy_static::lazy_static;
-use std::sync::Arc;
-use tokio::sync::Mutex;
 
 mod cell;
 mod cell_name;
@@ -45,10 +42,6 @@ mod cell_name;
 mod cells;
 mod cgroups;
 mod error;
-
-lazy_static! {
-    pub static ref CELLS: Arc<Mutex<Cells>> = Default::default();
-}
 
 #[derive(Debug, Clone)]
 pub struct CellSpec {
--- a/auraed/src/signal_handlers.rs
+++ b/auraed/src/signal_handlers.rs
@@ -28,21 +28,21 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
+use crate::runtime::CellService;
 use tokio::signal::unix::SignalKind;
+use tracing::error;
 
 /// Waits for a [SIGTERM] signal and...
-/// * Calls [Cells::broadcast_free]
-/// * Calls [Cells::broadcast_kill]
+/// * Calls [CellService::free_all]
 /// ---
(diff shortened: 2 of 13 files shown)


auraed/src/runtime/cell_service/cells/mod.rs
```diff
@@ -43,6 +46,10 @@ mod cells;
 mod cgroups;
 mod error;
 
+lazy_static! {
+    pub static ref CELLS: Arc<Mutex<Cells>> = Default::default();
```
dmah42, line 50: it feels odd to need this to be pub and static.  essentially, it's now a global (fine i guess but i would have thought some ownership of this thing would be good) but _also_ public.  can we at least limit it to the crate?

auraed/src/signal_handlers.rs
```diff
@@ -0,0 +1,48 @@
...
+/// * Calls [Cells::broadcast_free]
+/// * Calls [Cells::broadcast_kill]
+/// ---
+/// Returns after processing the first received signal.
+pub async fn terminate() {
+    let mut stream = tokio::signal::unix::signal(SignalKind::terminate())
+        .expect("failed to listen for SIGTERM");
+
+    let _ = stream.recv().await;
+    let mut cells = crate::runtime::CELLS.lock().await;
```
dmah42, line 43: should we instead send this signal to the cell service, which can handle the broadcast and kill calls on our behalf?  this means exposing the cell service instance to the handler rather than the implementation detail of the cells cache.

auraed/src/runtime/cell_service/cells/mod.rs
future-highway: On my phone so going off memory. 

I don't think I exposed it all the way out, but I'll make the change either way, so it doesn't get accidentally introduced.

auraed/src/signal_handlers.rs
future-highway: I think this will mean we put cells back on CellService (get rid of the global variable). Make a clone of CellService (cheap since it's only ARCs in the struct) and pass that into each signal handler that we make (or make a signal handler struct). 

Then have additional broadcast methods on CellService. 

Or do you mean making a rpc call? (I'm not sure how that would look)

Either way, I think we would be unnecessarily going through CellService which I look at as a hopefully slim/dumb-ish handler for the grpc api

dmah42: i was hoping we'd have some instance of the CellService we could get at from the signal handler.  is there no way to make _that_ the global thing?  or get a reference to it from tokio?

i didn't mean make an RPC from inside the binary, no.  definitely not from a signal handler!

future-highway: Got it. I'll do this

dmah42: \o/ no globals!

dmah42 (approved):

PR #224, opened 2023-01-02
Rename gRPC packages to follow naming best practices

RRethy:
See https://developers.google.com/protocol-buffers/docs/style#packages

The package names weren't unique nor did they have anything based on the project name. The renames were as follows:

```
observe.v0 -> aurae.observe.v0
runtime.v0 -> aurae.runtime.v0
schedule.v0 -> aurae.schedule.v0
```

### DIFF
--- a/api/v0/observe.proto
+++ b/api/v0/observe.proto
@@ -30,7 +30,7 @@
 
 syntax = "proto3";
 
-package observe.v0;
+package aurae.observe.v0;
 
 option go_package = "github.com/aurae-runtime/client-go/pkg/api/v0/observe";
 
--- a/api/v0/runtime.proto
+++ b/api/v0/runtime.proto
@@ -29,7 +29,9 @@
 \* -------------------------------------------------------------------------- */
 
 syntax = "proto3";
-package runtime.v0;
+
+package aurae.runtime.v0;
+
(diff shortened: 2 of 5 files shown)


dmah42 (dismissed):

RRethy:
Rebased on latest main since there were conflicts in the docs. Only change needed was to rerun `make docs`.

dmah42:
we no longer have generated files at HEAD so you'll need to rebase again, sorry.

but i think there's an issue where we have the version in the wrong place now.. i think the version needs to be at the outer layer so we don't end up with different modules having different versions (i made a similar comment elsewhere).

RRethy:
> we no longer have generated files at HEAD so you'll need to rebase again, sorry.

no worries lol, rebased and pushed, changes are significantly smaller.

> but i think there's an issue where we have the version in the wrong place now.. i think the version needs to be at the outer layer so we don't end up with different modules having different versions (i made a similar comment elsewhere).

I'm not sure what you mean here, do you mean `aurae.observe.v0` should end up being `aurae.v0.observe`? If so I agree based on our convo [here](https://github.com/aurae-runtime/aurae/pull/226#issuecomment-1369008159). That being said, I'd rather not include that in this PR.

dmah42 (approved):

PR #226, opened 2023-01-02
Restructure proto files

RRethy:
The previous structure was incompatible with the gRPC Go plugin for the buf tool since the generated Go code would get generated into a single directory with each file having a different package.

```
api/v0/observe.proto -> api/v0/observe/observe.proto
api/v0/schedule.proto -> api/v0/schedule/schedule.proto
api/v0/runtime.proto -> api/v0/runtime/runtime.proto
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ stdlibdocs:
 	$(error "No /usr/local/bin/protoc-gen-doc, install from https://github.com/pseudomuto/protoc-gen-doc")
 else
 stdlibdocs:
-	protoc --plugin=/usr/local/bin/protoc-gen-doc -I api/v0 --doc_out=docs/stdlib/v0 --doc_opt=markdown,index.md:Ignore* api/v0/*.proto
+	protoc --plugin=/usr/local/bin/protoc-gen-doc -I api/v0/runtime -I api/v0/schedule -I api/v0/observe --doc_out=docs/stdlib/v0 --doc_opt=markdown,index.md:Ignore* api/v0/*/*.proto
 endif
 
 crate: ## Build the crate (documentation)
--- a/api/v0/observe/observe.proto
+++ b/api/v0/observe/observe.proto
@@ -32,8 +32,6 @@ syntax = "proto3";
 
 package aurae.observe.v0;
 
-option go_package = "github.com/aurae-runtime/client-go/pkg/api/v0/observe";
-
 enum LogChannelType {
   LOG_CHANNEL_TYPE_UNSPECIFIED = 0;
   LOG_CHANNEL_TYPE_STDOUT = 1;
--- a/api/v0/runtime/runtime.proto
+++ b/api/v0/runtime/runtime.proto
@@ -32,8 +32,6 @@ syntax = "proto3";
 
 package aurae.runtime.v0;
 
-option go_package = "github.com/aurae-runtime/client-go/pkg/api/v0/runtime";
-
 /// Runtime
 /// ===
 ///
--- a/api/v0/schedule/schedule.proto
+++ b/api/v0/schedule/schedule.proto
@@ -32,6 +32,4 @@ syntax = "proto3";
 
 package aurae.schedule.v0;
 
-option go_package = "github.com/aurae-runtime/client-go/pkg/api/v0/schedule";
-
 // TODO Schedule subsystem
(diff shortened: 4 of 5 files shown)


dmah42:
i don't think we should do the separation of runtime into separate APIs.  i'm not sure it _does_ simplify future client code because we don't know what the future client is going to do.

i'm actually not even sure we should be separating the observer, runtime, schedule, etc APIs _either_.. ie, it should maybe be `aurae.v0.runtime` with everything under `aurae` rather than `aurae.runtime.v0`.

either way, this suggests an independence of API that is not necessarily accurate.

RRethy:
> i don't think we should do the separation of runtime into separate APIs. i'm not sure it does simplify future client code because we don't know what the future client is going to do.

I have a built client library which is where I encountered the package issues that lead me to these changes, the full client can be seen at https://github.com/RRethy/aurae/tree/full-go-client/client-go.

> i'm actually not even sure we should be separating the observer, runtime, schedule, etc APIs either.. ie, it should maybe be aurae.v0.runtime with everything under aurae rather than aurae.runtime.v0.

Do you mean we would have a single gRPC service in runtime.proto or multiple services in runtime.proto?

I do see what you mean about the API not being independent, but I'm not a fan of having multiple services restricted to the same version, at that point it would indicate to me it should be a single service which is what I'm gathering from what you're saying (which I'd be in favour of).

An alternative fix to the packaging issue would be to adopt a structure of `api/v0/runtime/runtime.proto`, `api/v0/observe/observe.proto`, `api/v0/schedule/schedule.proto`. This would keep the existing structure but would fix the buf Go code generation, but I'd lean towards merging into a single service if the versions need to stay consistent on the endpoints.

dmah42:
> > i don't think we should do the separation of runtime into separate APIs. i'm not sure it does simplify future client code because we don't know what the future client is going to do.
> 
> I have a built client library which is where I encountered the package issues that lead me to these changes, the full client can be seen at https://github.com/RRethy/aurae/tree/full-go-client/client-go.

👍🏻 

> 
> > i'm actually not even sure we should be separating the observer, runtime, schedule, etc APIs either.. ie, it should maybe be aurae.v0.runtime with everything under aurae rather than aurae.runtime.v0.
> 
> Do you mean we would have a single gRPC service in runtime.proto or multiple services in runtime.proto?

what we currently have: multiple services in "runtime" that are versioned together.  (i'd even go further and consider a single aurae API that is versioned together but i think that bird has flown, and i'm not confident that the client code would be easy to use).

> 
> I do see what you mean about the API not being independent, but I'm not a fan of having multiple services restricted to the same version, at that point it would indicate to me it should be a single service which is what I'm gathering from what you're saying (which I'd be in favour of).

if we did (and i think i am arguing myself around to it) we'd have the aurae RuntimeService with Cell, Pod, and Executable endpoints.  i think we didn't go that way because it is much easier to conceive that "i'm working with Cells right now" and only connect to the CellService.

i don't have a strong opinion here, but i do like where we currently have landed... the issue (which might already be an issue to be honest) is the independent versioning.  i don't want to try to explain why observe v0 doesn't work with runtime v1 to someone so encoding the dependency in the global version makes the most sense to me.

> 
> An alternative fix to the packaging issue would be to adopt a structure of `api/v0/runtime/runtime.proto`, `api/v0/observe/observe.proto`, `api/v0/schedule/schedule.proto`. This would keep the existing structure but would fix the buf Go code generation, but I'd lean towards merging into a single service if the versions need to stay consistent on the endpoints.

that would be a good iteration from what we have that is less controversial :)

krisnova:
Can we get this PR updated and these conflicts resolved please?

RRethy:
> i don't want to try to explain why observe v0 doesn't work with runtime v1 to someone so encoding the dependency in the global version makes the most sense to me.

Yea I see what you mean for versioning.

>> An alternative fix to the packaging issue would be to adopt a structure of api/v0/runtime/runtime.proto, api/v0/observe/observe.proto, api/v0/schedule/schedule.proto. This would keep the existing structure but would fix the buf Go code generation, but I'd lean towards merging into a single service if the versions need to stay consistent on the endpoints.

> that would be a good iteration from what we have that is less controversial :)

Updated the PR for these changes, although I'm alright with closing the PR since this was intended to fix an issue with how buf generated the Go code for a Go client in aurae-runtime/aurae, but this is being working on by @muffinman so his approach may not encounter the same issues.

krisnova:
Please see #231 and let's discuss if we can (or should?) adopt the hurricane naming pattern here?

krisnova:
These files are still conflicting

krisnova (approved):

dmah42 (approved):

PR #227, opened 2023-01-02
Add initial go client

RRethy:
Follow-up to https://github.com/aurae-runtime/aurae/pull/226 so the first 16 commits are duplicated from that branch, unfortunately GitHub won't let me use that PR as the base since it's not under aurae-runtime.

This PR adds an initial Go client under client-go. It only has code generate by buf, a future PR (probably some time tmrw or tuesday) will have a more fleshed out client.

The PR boils down to adding this to `buf.gen.yaml`, and running `make proto`.

```yaml
  - plugin: buf.build/protocolbuffers/go:v1.28.1
    out: client-go/pkg/api
    opt: paths=source_relative
  - plugin: buf.build/grpc/go:v1.2.0
    out: client-go/pkg/api
    opt: paths=source_relative
```


dmah42:
meta comment: I don't think we should be committing generated code.

api/runtime/cell/v0/cell.proto
```diff
@@ -29,68 +29,10 @@
 \* -------------------------------------------------------------------------- */
```
future-highway, line 29: I like that the services are broken into smaller files. Should there be a higher level runtime file? If we have shared messages, where would we put those?

dmah42: the issue i have here is that each broken down part now has its own version, which i don't think is what we want.. we want aurae.<version>.<service> (where service is runtime or runtime.cells, or whatever).

having aurae.<service>.<version> implies that we might have runtime.cells.v1 and observe.v0 which i would like to avoid.

client-go/go.mod
```diff
@@ -0,0 +1,16 @@
+module github.com/aurae-runtime/aurae/client-go
```
future-highway, line 1: I thought the go code would be in the ae repo. It doesn't matter to me; I'm just checking that the contributors working on it (I know that includes you) are in agreement that this is the right decision at this point in time.

Side note: @krisnova @dmah42 if we make this repo have more general code, maybe we should consider moving all the rust crates into the crates directory.

crates/aurae-client/src/runtime/cell_service.rs
```diff
@@ -28,10 +28,10 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
-use aurae_proto::runtime::*;
+use aurae_proto::runtime::cell::*;
 
 macros::ops_generator!(
-    runtime,
+    cell,
```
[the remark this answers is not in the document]
future-highway: Realizing I got this mixed up with auraescript's macro. I still suspect there will be an issue in the generated code. I'll have to pull down the PR to check.

auraescript/gen/cell.ts
```diff
@@ -705,3 +452,24 @@ export type Exact<P, I extends P> = P extends Builtin ? P
 function isSet(value: any): boolean {
   return value !== null && value !== undefined;
 }
+export class CellServiceClient implements CellService {
+allocate(request: CellServiceAllocateRequest): Promise<CellServiceAllocateResponse> {
+    // @ts-ignore
+    return Deno.core.ops.ae__cell__cell_service__allocate(request);
```
future-highway, line 458: See my comment incorrectly placed on the aurae-client code

client-go/go.mod
dmah42: yeah i had expected this to be in a separate repo. i didn't notice this.

auraescript/macros/src/ops.rs
```diff
@@ -166,7 +166,7 @@ fn typescript_generator(input: &OpsGeneratorInput) {
 
     let ts_path = {
         let mut out_dir = gen_dir.clone();
-        out_dir.push(format!("v0/{module}.ts"));
+        out_dir.push(format!("runtime/{module}/v0/{module}.ts"));
```
future-highway, line 169: This isn't the correct way to fix the macro as we have more modules than `runtime`, but if it works in this PR and you are not comfortable making the change, I can fix it after if we merge this as is.

api/runtime/cell/v0/cell.proto
future-highway: Definitely agree with the version being in the wrong spot. I'm also inclined to agree with your comment overall ([more context](https://github.com/aurae-runtime/aurae/pull/226#issuecomment-1368728945)).

My only reason for liking the multiple files is that smaller files are easier to grasp, but that is not a good reason to break things up if it causes other issues.

RRethy: I posted a follow-up on the structure, https://github.com/aurae-runtime/aurae/pull/226#issuecomment-1368992622.

client-go/go.mod
RRethy: What's the rational behind putting it in a different repo?

My logic for putting it in aurae-runtime/aurae was the following:

- Keep the source of truth in the same repo to avoid getting out sync
- The rust and typescript generated code are in this repo as well
- The proto files will have a `option go_package = "..."` which points to it's own repo instead of an external repo that can easily break
- If we want to move the code into another repo, we could follow kubernetes/client-go which just syncs that code from https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/client-go which acts as the source of truth (and lives alongside the types it exposes)

If you all don't want it here though I can close the PR and we can figure out a more appropriate place for it.

RRethy:
> meta comment: I don't think we should be committing generated code.

Would this apply to the rust/typescript generate code?

I've usually seen gRPC generate code being an exception to that rule. The issue I'd see with not committing the generated code is that if we provide a client wrapper around the generated code (similar to https://github.com/aurae-runtime/client-go/blob/main/client.go), it would break downloading the code with `go get .../client-go` since the code wouldn't be generated for downstream users.

dmah42:
> > meta comment: I don't think we should be committing generated code.
> 
> Would this apply to the rust/typescript generate code?
> 
> I've usually seen gRPC generate code being an exception to that rule. The issue I'd see with not committing the generated code is that if we provide a client wrapper around the generated code (similar to https://github.com/aurae-runtime/client-go/blob/main/client.go), it would break downloading the code with `go get .../client-go` since the code wouldn't be generated for downstream users.

it does apply to that, yes.  in fact we don't have that generated code at head any more.

perhaps this removal was a mistake.. i wasn't aware that go required the generation to be checked in.  that's less than ideal (it makes for noisy commits, merge issues, and sometimes people forget to generate the code before committing).

unless we have some github workflow that generates the code and checks it in instead, but that's an education problem.

future-highway:
> > meta comment: I don't think we should be committing generated code.
> 
> Would this apply to the rust/typescript generate code?

Just answering ^^^

Yes, I would expect it to apply to all generated code. The only benefit I'm seeing from checking in generated code is that I can see that changes in the generators are outputting the expected result. However, that can't even be trusted. How can we be sure a PR generated the code as the last change before pushing? I think I've made the mistake before where I didn't include the changes in the generated code in the PR. If someone pulled the source at that point, the generated code would be correct only after running the build.

I don't know/haven't considered if an exception is warranted.

krisnova:
Are we putting Go code in this repository now? 

I thought we had decided to put the `client-go` [here](https://github.com/aurae-runtime/ae/tree/main/client#client-go) as per the https://github.com/aurae-runtime/ae/issues/13#issuecomment-1368072010.

RRethy:
> Are we putting Go code in this repository now?
> 
> I thought we had decided to put the `client-go` [here](https://github.com/aurae-runtime/ae/tree/main/client#client-go) as per the [aurae-runtime/ae#13 (comment)](https://github.com/aurae-runtime/ae/issues/13#issuecomment-1368072010).

Ah I didn't see that issue, I'll close this and move discussion into that issue.

krisnova:
@RRethy please mention either in the Discord or on the issue that you intend on working on it. I know other people are also considering this.

Issue #197, opened 2022-12-27
aurae/api/v0/ runtime.proto: use repeated fields for stringly-typed fields containing multiple items?


matttproud:
I was looking over https://github.com/aurae-runtime/aurae/blob/main/api/v0/runtime.proto, and two instances of this caught my attention.

>   /// A comma-separated list of CPU IDs where the task in the control group
>  /// can run. Dashes between numbers indicate ranges.

Would it be prudent to represent [`Cell.cpu_cpus`](https://github.com/aurae-runtime/aurae/blob/main/api/v0/runtime.proto#L115) and [`Cell.cpu_mems`](https://github.com/aurae-runtime/aurae/blob/main/api/v0/runtime.proto#L131) as [`repeated`](https://developers.google.com/protocol-buffers/docs/proto3#specifying_field_rules) fields so that some of the stringly-typed input can be retired in favor of a structured data schema?

So the fields from `Cell` would look something like this:

```
repeated string cpu_cpus = 2;
...
repeated string cpu_mems = 4;
```

For the expression of ranges, I might consider using the [`oneof`](https://developers.google.com/protocol-buffers/docs/proto3#oneof) capability.  A sketch:

```
message CPUSpec {
  message Range {
    string start = 1;
    string end = 2;
  }
  oneof spec {
    string id = 1;
    Range range = 2;
  }
}

repeated CPUSpec cpu_cpus = 2;
```

dmah42:
this was discussed in the discord and we decided to keep it like this so it's close to the underlying API.

I agree repeated numbers might be clearer and more type safe but the format is so specific we didn't want to take things too far from that.

matttproud:
I see.  Would you want to add a note to the fields to indicate that the
values should be provided as literals according to a format dictated by an
external system?

future-highway:
I agree with @dmah42 that we should mimic the underlying API, however, I was thinking of ways to have both. 

We could have, for example:
```proto
message Cell {
  ...  
  oneof cpu_spec {
    string cpu_spec_native = 3;
    CPUSpecRange cpu_spec_range = 4;
  }
  ...
```

where native would match the underlying api, but the other choices would provide more typed options that we would convert to the native version. And we could always add more formats and it won't be a breaking change.

dmah42:
i would prefer to have one way to define something in the API, just for simplicity.

[...]

PR #228, opened 2023-01-02
RFC: use repeated ints instead of native format in the API

dmah42:
Implementation of #197.

Created to source comments on whether this is what we want: a more natural proto API that doesn't match the native cgroup API.


api/v0/runtime.proto
```diff
@@ -151,8 +151,8 @@ message Cell {
   uint64 cpu_shares = 3;
 
   /// Same syntax as the cpus field of this structure, but applies to
-  ///  memory nodes instead of processors.
-  string cpu_mems = 4;
+  /// memory nodes instead of processors.
+  repeated int32 cpu_mems = 4;
```
future-highway, line 155: I don't like that we are moving away from the standard format. I think we should provide StandardFormat+. I'll make a more clear comment on my thoughts in a top-level comment

auraed/src/runtime/cell_service/cells/cgroups/cpu_cpus.rs
```diff
@@ -1,29 +1,34 @@
...
+    pub fn new(cpu_cpus: Vec<i32>) -> Self {
         Self(cpu_cpus)
     }
 
-    pub fn into_inner(self) -> String {
+    pub fn into_inner(self) -> Vec<i32> {
         self.0
     }
+
+    pub fn as_string(&self) -> String {
```
future-highway, line 24: This should be implemented using the Display trait which gives a `to_string`: https://doc.rust-lang.org/std/string/trait.ToString.html

dmah42: i deliberately didn't because this is not for display; it's a conversion from one representation to another.  if it's idiomatic to use Display for this then i can, but i didn't think it was the right fit.

future-highway:
Expanding on my review comment regarding "StandardFormat+" api. To do that, I think we need multiple changes...

To get a more standard like api, I think we need to break up the messages. Assuming [these docs](https://docs.kernel.org/admin-guide/cgroup-v2.html) are the standard for cgroups, I found them by googling "cgroup cpu format" and if I scroll down, I find a section [Cpuset Interface Files](https://docs.kernel.org/admin-guide/cgroup-v2.html#cpuset-interface-files), with familiar, but not exactly the same names. I think we should be using the exact same structure and names (with the exception of capitalization format, which should match idiomatic proto).

So, IMO, we should have something like:
```proto
message Cell {
  ...
  CpuSet cpu_set = 2;
  ...
}

message CpuSet {
  string cpus = 1;
  string mems = 2;
}
```
which would be based on the standard format.

Now, to enhance it, we could change it to:
```proto
message CpuSet {
  oneof cpus_format { 
    string cpus = 1;
    Int32List cpus_as_list = 10;
  }

  oneof mems_format {
    string mems = 2;
    Int32List mems_as_list = 11;
  }
}

message Int32List {
  repeated int32 items = 1;
}
```
which has no breaking changes from the first message, and we can continue to expand over time with, for example, `cpus_as_range`.

By using the standard naming, it makes it much easier to look up the documentation for what the native format is (in case our documentation is lacking), and get an explanation of what setting the field does. By using `oneof` we've provided and enhancement to the usability of the system.

*I may not have gotten the implementations correct in these examples, but my overall point is that we should model the underlying api, and then add enhancements that make it more type safe and easier to use.*

dmah42:
i like the idea of following the API more closely with the nested CpuSet message.

i don't like using a oneof to provide two different APIs.  there should be one way to interact with the API which is what we should decide here.

argument for the replacement API (as_list): a programmatic client will have an easier time building the list and using the API.   say, for instance, a scheduler that is managing resources on hosts.  if we offer the string API, that client (and all other clients) have to form the right format when interacting with our API.  with the as_list API, we have taken the burden from them and given them a simpler view into the underlying API.

future-highway:
> i like the idea of following the API more closely with the nested CpuSet message.

Yeah, I think it's a good idea to run with this and also have messages for all the cgroup controllers that we want to expose. 

> i don't like using a oneof to provide two different APIs. there should be one way to interact with the API which is what we should decide here.

I guess our difference stems from me (maybe incorrectly) not considering it multiple APIs. I think if we can pick one to start, and later weight any demand for a more flexible API, but if we use the shorter field names for an opinionated format right now, we will either need to change the names (I know it's not technically considered breaking in proto, but definitely annoying) or live with the potential of maybe less intuitive naming.

> argument for the replacement API (as_list): a programmatic client will have an easier time building the list and using the API. say, for instance, a scheduler that is managing resources on hosts. if we offer the string API, that client (and all other clients) have to form the right format when interacting with our API. with the as_list API, we have taken the burden from them and given them a simpler view into the underlying API.

An additional, and similar argument, is that the as_list API is less prone to human error given the strict typing and format. 

Going in an opinionated way (as opposed to using the underlying format) means we will need to make format decisions for all the fields (may not be a big deal). I'm more wondering if all the functionality possibilities that we want to expose will be able to be expressed in the format we decide on.

auraed/src/runtime/cell_service/cells/cgroups/cpu_cpus.rs
future-highway: I would probably name it something else then. Maybe `as_controller_format`? Or maybe it should be `to_controller_format` ([guidelines](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv))?

RRethy: Alternatively you could use a type alias, e.g. CpusStr, and implement the `From` trait so the conversion can be done with `.into()`.

krisnova:
Pulse check? Where are we at with this one?

dmah42:
> Pulse check? Where are we at with this one?

I feel pretty strongly that we should only have one way to represent the field in the API.

I have a weakly held opinion that the repeated int is a more natural representation and easier to validate, but I can see that familiarity with the underlying API might lean us towards consistency with it.

I think we need an executive decision :D

then I can finalise this PR or close it.

krisnova:
We want to prefer flexibility over structure at this point. I believe we want to go with the `oneof` approach. 

The reason for this is that we want to "abstract" and "bring up" the lower level functionality to a more friendly/supportive layer of the stack. One of my biggest gripes with Kubernetes APIs is having to spend way too much time in the weeds trying to figure out how to express a fairly basic concept. I will be a happy person the day I can pass 

```
a,b,c
a , b , c
a b c
abc 
bac 
c, a, b
```

to a function and it spit out the same results. 

I think @future-highway's suggestion is the way to go. If we want to change `CpuSet` in the future we can, and this way we have some flexibility moving forward. I don't want to live in a world where we are cutting a new version of the API and breaking everybody's shit because we decided to change a repeated int32 to a repeated int64 in the future. This gives us a way around that. Strong +1 on the `oneof` tactic.

As for the types such as `Int32List` I would stay away from re-creating a basic set of repeated types. However in the case where we have special conditions such as "Range" I would have a type for that. Maybe `Int32Range` or similar. 

tldr:

> we should model the underlying api, and then add enhancements that make it more type safe and easier to use.

Yes. Decision made. Leverage the `oneof` tactic, and get rid of the `Int32List` but feel free to implement `Int32Range` (the same goes for all types: `Int64Range`, etc)

krisnova (changes requested):
See my comment on `CpuSet` and `Int32List` and `oneof`.

Please make necessary changes.

future-highway:
> Yes. Decision made. Leverage the `oneof` tactic, and get rid of the `Int32List` but feel free to implement `Int32Range` (the same goes for all types: `Int64Range`, etc)

Regarding `Int32List`, we need something like that because I don't think you can have a repeated field in a `oneof` so a wrapper is needed.

krisnova:
> Regarding Int32List, we need something like that because I don't think you can have a repeated field in a oneof so a wrapper is needed.

Then lets document this standard [in the convention in api/README.md](https://github.com/aurae-runtime/aurae/tree/main/api#api-definition-convention) and make it so.

dmah42:
> We want to prefer flexibility over structure at this point. I believe we want to go with the `oneof` approach.

I strongly disagree.

Once we have oneof in the API it's extremely difficult to migrate away from it.  Proto APIs are great, but migration paths stink.

If there's not enough for the repeated int approach to define the CPU set then we should just drop this and continue with the string _until_ there's support for repeated int at which point we should migrate _once_.

> 
> The reason for this is that we want to "abstract" and "bring up" the lower level functionality to a more friendly/supportive layer of the stack. One of my biggest gripes with Kubernetes APIs is having to spend way too much time in the weeds trying to figure out how to express a fairly basic concept. I will be a happy person the day I can pass
> 
> ```
> a,b,c
> a , b , c
> a b c
> abc 
> bac 
> c, a, b
> ```
> 
> to a function and it spit out the same results.
> 
> I think @future-highway's suggestion is the way to go. If we want to change `CpuSet` in the future we can, and this way we have some flexibility moving forward. I don't want to live in a world where we are cutting a new version of the API and breaking everybody's shit because we decided to change a repeated int32 to a repeated int64 in the future. This gives us a way around that. Strong +1 on the `oneof` tactic.
> 
> As for the types such as `Int32List` I would stay away from re-creating a basic set of repeated types. However in the case where we have special conditions such as "Range" I would have a type for that. Maybe `Int32Range` or similar.
> 
> tldr:
> 
> > we should model the underlying api, and then add enhancements that make it more type safe and easier to use.
> 
> Yes. Decision made. Leverage the `oneof` tactic, and get rid of the `Int32List` but feel free to implement `Int32Range` (the same goes for all types: `Int64Range`, etc)

future-highway:
I don't want to push the `oneof` option too much. We can always add a `oneof` later if the want/need arises. It shouldn't be a breaking change to wrap an existing field in a `oneof`.

I do want to reiterate that I think we should build our messages to be more closely aligned with the underlying API.

krisnova:
Are we blocked on anything here? I keep seeing this in the queue and am wondering what I haven forgotten to do?

dmah42:
it's on me to document a decision.

krisnova:
boops -- should we close this?

dmah42:
sorry I still owe a decision in the documentation

PR #229, opened 2023-01-02
remove generated files from being committed to the repo

dmah42:
I think this is safe and have confirmed that the files are regenerated on `make`

The proto generation wasn't a dependency for making docs but should have been.


future-highway (approved):
I really like this change!

PR #232, opened 2023-01-02
Make `Executable` async using `tokio::process::Command`

future-highway:
`Executable` now uses `tokio::process::Command` for its async abilities. It no longer uses the `Process` enum that we have in the module as it seems like it's more trouble then it's worth.

`NestedAuraed` is still synchronous as I would rather see how that evolves to fill our needs prior to making it async. It also still uses `Process`, but there is probably no reason for `Process` to exist separately so it will probably be incorporated into `NestedAuraed`.

I also refactored LogChannel into a more expected structure.

### DIFF
--- a/auraed/src/logging/log_channel.rs
+++ b/auraed/src/logging/log_channel.rs
@@ -31,64 +31,40 @@
 // @todo @krisnova remove this once logging is futher along
 #![allow(dead_code)]
 //
+use super::get_timestamp_sec;
 use aurae_proto::observe::LogItem;
 use tokio::sync::broadcast::{self, Receiver, Sender};
-use tracing::error;
-
-use super::get_timestamp_sec;
 
 /// Abstraction Layer for one log generating entity
 /// LogChannel provides channels between Log producers and log consumers
-#[derive(Debug)]
-pub struct LogChannel {
-    producer: Sender<LogItem>,
-    name: String,
+#[derive(Clone, Debug)]
+pub(crate) struct LogChannel {
+    pub name: String,
+    tx: Sender<LogItem>,
 }
 
 impl LogChannel {
     /// Constructor creating the channel for log communication
     pub fn new(name: String) -> LogChannel {
         // TODO: decide for a cap. 40 is arbitrary
-        let (producer, _) = broadcast::channel(40);
-        LogChannel { producer, name }
-    }
-
-    /// Getter for the name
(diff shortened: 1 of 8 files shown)


auraed/src/logging/log_channel.rs
```diff
@@ -31,65 +31,41 @@
...
+        self.tx.subscribe()
     }
 
     /// Wrapper that sends a log line to the channel
-    pub fn log_line(producer: Sender<LogItem>, line: String) {
+    pub fn send(&self, line: String) {
         // send returns an Err if there are no receivers. We ignore that.
-        let _ = producer.send(LogItem {
+        let _ = self.tx.send(LogItem {
             channel: "unknown".to_string(),
```
future-highway, line 63: Should this be the LogChannel's name?

krisnova (dismissed):

auraed/src/logging/log_channel.rs
dmah42: yes!

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -47,168 +39,89 @@ impl Executable {
...
-        let process = exec(command)?;
+        let mut child = command
+            .current_dir("/")
+            .stdout(Stdio::piped())
+            .stderr(Stdio::piped())
+            .spawn()?;
+
+        let stdout = child.stdout.take().expect("stdout");
+        let log_channel = LogChannel::new(format!("{}::stdout", self.name));
+        let span = info_span!("running process", name = ?self.name);
```
dmah42, line 60: I think this should be inside the tokio spawn thread?

dmah42 (approved):
I'm curious if this resolves the delay on stop.

auraed/src/runtime/cell_service/executables/executable.rs
dmah42: ah gotcha. you're right, of course.  i keep forgetting the extra "entered" call that's necessary to enable the span.

dmah42 (approved):
<3 this change

future-highway:
> I'm curious if this resolves the delay on stop.

I think yes and no. Assuming the sleep call was the cause, then we would no longer have an "unnecessary" delay. However, this should read stdout/err to the end, so it will take whatever time that takes for `kill` to return.

PR #235, opened 2023-01-05
Introduce image and original pod plumbing

krisnova:
Solve image pulling and OCI initial example (OCI bundles, and archive, etc)

This is not a functionally complete PR as this is merely introducing the initial scaffolding to the Pod service. We will continue to build the service in subsequent smaller PRs. The main APIs are all currently marked with a `todo!()` to break the client if they happened to be called.


auraed/src/runtime/pod_service/pod_service.rs
```diff
@@ -60,14 +67,44 @@ impl pod_service_server::PodService for PodService {
...
+        let pod = request.pod.expect("pod");
+        let name = pod.name;
+        let image = pod.image;
+        info!(
+            "PodService: allocate() name={:?} image={:?}",
+            name.clone(),
+            image.clone()
+        );
+
+        let _container_bundle = bundle::container::container(&image);
```
krisnova, line 80: @future-highway here is where we need to turn an "image string" into an "OCI bundle" on the local filesystem. 

After the OCI bundle is "decompressed" locally we can then use the files in the bundle to build the container below.

krisnova:
Hey future just a heads up I am rebasing and moving forward a little bit

auraed/src/lib.rs
```diff
@@ -138,7 +146,6 @@ pub async fn daemon() -> i32 {
     init::init(options.verbose, options.nested).await;
 
     info!("Starting Aurae Daemon Runtime");
-    info!("Options: {options:#?}");
```
future-highway, line 141: Did we not want this at least as `trace!()`

auraed/src/runtime/pod_service/pod_service.rs
```diff
@@ -31,27 +31,33 @@
 #![allow(unused)]
 #![allow(clippy::module_inception)]
 
+use anyhow::{Context, Result};
```
future-highway, line 34: I'm not spotting where this file is even using `anyhow`, but we should refactor it out in a future PR

krisnova:
Read for review/merge 😄

future-highway (approved):
I didn't pull it to try anything, and my comments are really just notes for the future iterations.

PR #236, opened 2023-01-05
Cells and Executable Security and Unsharing

future-highway:
This pull request was started by @future-highway and finished by @krisnova.

This change adds the following API flags to the runtime cell service:

```proto

  /// Will isolate the process (and proc filesystem) from the host.
  /// Will unshare the pid, ipc, uts, and mount namespaces.
  /// The cgroup namespace is always unshared with the host.
  ///
  /// Default: false
  bool isolate_process = 10;

  /// Will isolate the network from the host.
  /// Will unshare the net namespaces.
  /// The cgroup namespace is always unshared with the host.
  ///
  /// Default: false
  bool isolate_network = 11;

```

These are minimal isolation features that can optionally be enabled for "extra" isolation at the cell level.

Anything more granular or stronger than this is reserved for the pod service as it can be allocated and started using a traditional container paradigm.

To put it simply the **Isolate Process** flag will isolate the mount, pid, uts, and ipc namespaces and manage any of the forking and proc mounting that needs to happen to render the desired effects.

The **Isolate Network** flag is much simpler and just cordons off a network for the cell away from the host. The only network device remaining after the isolation is loopback. The route table is empty.

### @future-highway's original content

> 
> None of this really works as it should yet. I think only unsharing `pid` works as it should, if `mount` is also unshared. But, I tried to make notes with what is happening and what is going wrong, so hopefully those are accurate.
> 
> Putting this up now in the event that either:
> * I'm on the wrong track (aka don't keep spending time on it)
> * I'm on the right track, and maybe someone knows how to do this easily and wants to jump in, or maybe can solve some of it
> 
> **CAUTION**: Sometimes things break in my vm. After stoping the auraed process I run:
> ```
> sudo -E kill -9 $(cat /sys/fs/cgroup/ae*/_/cgroup.procs); pkill -9 -f auraed; rmdir /sys/fs/cgroup/ae-*/*; rmdir /sys/fs/cgroup/ae-*; umount /ae-*/*; umount /ae-*; rmdir /ae-*/*; rmdir /ae-*; rm -rf /ae-*; rmdir /ae-*/*; rmdir /ae-*;
> ```
> 
> (sometimes I need to run it 2x)
> 
> *While some of that command was written by Nóva, I expanded it to include other things. Potentially dangerous and dumb things. Please read it over. I don't know what I'm doing and always have the option to reset my vm.*


dmah42:
https://github.com/apache/mesos/blob/96339efb53f7cdf1126ead7755d2b83b435e3263/src/slave/containerizer/mesos/launch.cpp <-- this is how mesos launches stuff.  see `prepareMounts` and `enterChroot` for hopefully some useful hints.

`MesosContainerizerLaunch::execute` is the entry point iirc.

future-highway:
Update on issue with not being able to launch nested auraed post changing the root dir. Potential cause:

std's `Command` uses `libc::execvp`, and a [note](https://github.com/apache/mesos/blob/96339efb53f7cdf1126ead7755d2b83b435e3263/src/slave/containerizer/mesos/launch.cpp#L1177) in the mesos link above points out...
> execvpe and execvp will only search executable from the current working directory if environment variable PATH is not set.

The info is confirmed on the [man page](https://man7.org/linux/man-pages/man3/exec.3.html)
> These functions duplicate the actions of the shell in searching
       for an executable file if the specified filename does not contain
       a slash (/) character. The file is sought in the colon-separated
       list of directory pathnames specified in the PATH environment
       variable.  If this variable isn't defined, the path list defaults
       to a list that includes the directories returned by
       confstr(_CS_PATH) (which typically returns the value
       "/bin:/usr/bin") and possibly also the current working directory;
       see NOTES for further details.

krisnova:
@future-highway from the stream can we please:

 - Remove all of the namespace options from the `Cells` API other than **Pid** and **Net** for now?
 - Add a new bool flag in the API called something such as `MountProcInsecure` which will toggle mounting procfs in the cell?
 - Add a new bool flag in the API called something such as `MountDevInsecure` which will toggle mounting /dev in the cell?
 - Add a new bool flag in the API called something such as `MountSysInsecure` which will toggle mounting /sys in the cell?

Make sure to keep the clone3 work you have done please. I think most of these changes will be in the `shared_namespaces.rs` file.

future-highway:
@krisnova I don't think we need to change the api (other than maybe commenting out the unimplemented fields), if the current api is the goal. We can take the namespaces one at a time and get the api we want and if the flags you mentioned are closer to what we want, then we can work towards those.

However, your comment makes me think that the api is too simple (doesn't have enough config options). To use what seems like a simple namespace, `uts`, we can have more than just share/unshared. We can have (all or some of):
- shared (hostname & domainname always equals parent, and change together)
- unshared (hostname & domainname are inherited but change independently of parent)
- unshared (hostname & domainname are set to custom values and are independent of parent)

My initial feeling is that we should have more config, with good defaults. But we could also make some executive decisions and say, "this is how it works when you don't share namespace x". 

RE: `NET` & `PID`...

---

### `NET` 

Seems like it just works when it is not shared:

#### examining the parent...
```bash
# ip link
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: enp0s5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP mode DEFAULT group default qlen 1000
    link/ether 00:a1:ba:51:4c:12 brd ff:ff:ff:ff:ff:ff
```

#### examining the child...
```bash
# nester -a -t PID_FOR_NESTED_AURAED
# ip link
1: lo: <LOOPBACK> mtu 65536 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
```

---

### `PID`

This one is not as straight forward from what I've gathered. If we un-share the ns, we always get pid 1 for the nested auraed (good). However, `/proc` will always be the view of the process that mounted it (trouble). So... 

If the `mount` ns is not shared, we can do:
  1. Stop the child process's mounts from propagating to the parent using the flags `MS_SLAVE | MS_REC` (or `MS_PRIVATE | MS_REC`) on the '/'.
  2. Then mount `/proc` in the child and then the child will get the expected view of the system.

If the mount is shared, we can't mount `/proc` without changing the root of the child process, otherwise we will break the parent's mounts (i.e., the parent will see the system as if it were the child).

---

krisnova:
Just jumping in here. @future-highway and I have traded PRs. 

So I am not in charge of this illustrious work of art.

krisnova:
Ready for review :)

auraed/src/runtime/cell_service/executables/auraed/isolation_controls.rs
```diff
@@ -0,0 +1,126 @@
...
+        let root = PathBuf::from("/");
+        nix::mount::mount(
+            Some(&root),
+            &root,
+            None::<&str>, // ignored
+            nix::mount::MsFlags::MS_PRIVATE | nix::mount::MsFlags::MS_REC,
+            None::<&str>, // ignored
+        )
+        .map_err(|e| io::Error::from_raw_os_error(e as i32))?;
+        info!("Isolation: Bind mounted root dir (/) in cell");
```
dmah42, line 69: I think strictly speaking this is not a bind mount as the BIND flag is not set.

dmah42 (approved):

auraed/src/runtime/cell_service/executables/auraed/isolation_controls.rs
krisnova: I’ll update — I also have an erroneous println in the logger as well I need to clean up

auraed/src/runtime/cell_service/executables/auraed/isolation_controls.rs
```diff
@@ -0,0 +1,126 @@
...
+            None::<&str>, // ignored
+            nix::mount::MsFlags::MS_PRIVATE | nix::mount::MsFlags::MS_REC,
+            None::<&str>, // ignored
+        )
+        .map_err(|e| io::Error::from_raw_os_error(e as i32))?;
+        info!("Isolation: Mounted root dir (/) in cell");
+        Ok(())
+    }
+
+    pub fn isolate_process(
```
dmah42, line 73: maybe add some info logging as above to confirm that proc is mounted?

auraed/src/runtime/cell_service/executables/executable.rs
```diff
@@ -64,7 +64,8 @@ impl Executable {
             let mut stdout = BufReader::new(stdout).lines();
             while let Ok(Some(line)) = stdout.next_line().await {
                 let entered_span = span.take().expect("span").entered();
-                info!(level = "info", channel = log_channel.name, line);
+                //info!(level = "info", channel = log_channel.name, line);
```
dmah42, line 67: did you remove this on purpose? it's logging the spawned executable output.

dmah42 (approved):

auraed/src/runtime/cell_service/executables/executable.rs
krisnova: Its really not. It's "flooding" the entire output line by line with an enormous tracing line. To be honest it was obnoxious. Every line in the output would print a few hundred "meta" characters and then the content of the line. 

```
output 1
output 2
output 3
```

Was printed like:

```
runtime::cells::aurae::executable::output::something::meta::another {} log () line () date:: time -> output -> warn ::something::else -> output::runtime output 1
runtime::cells::aurae::executable::output::something::meta::another {} log () line () date:: time -> output -> warn ::something::else -> output::runtime output 2
runtime::cells::aurae::executable::output::something::meta::another {} log () line () date:: time -> output -> warn ::something::else -> output::runtime output 3
```

obviously I just made the log lines up, but you get the idea.

PR #240, opened 2023-01-06
Remove cached cargo deps from the docker image

jesse-peters:
We're getting some 403 errors on build pipelines when this image is pushing to GHCR. Heres my thoughts that triggered this PR to fix it:

1) We have a 10GB limit on the cache for this project. We're at our about that constantly because this cache holds the cargo files, as well as the docker cache
2) Github is nice and cleans this up for us by when we used it last. But theres not much runway for things to exist. The image is ~1GB and I think the cargo dependencies are another ~300MB. And each change to Cargo.lock across branches would trigger both some new docker layers, as well as a new cache of the .cargo folders
3) This build is attempting to use a layer that it thinks is cached somehow, but it cant find the actual blob when it trys to copy from it. But it carries on
4) When GHA trys to push the image, its referencing a layer that doesnt actually exist in the cache, so it fails 

I am removing the `cargo-chef` step in the dockerfile. This will remove the cargo cache from the image itself, and hopefully remediate this issue now, and going forward. Because the image being built now is just the system dependencies, it will not be rebuilt nearly as often.

### DIFF
--- a/.github/workflows/build-with-containers.yml
+++ b/.github/workflows/build-with-containers.yml
@@ -56,7 +56,7 @@ jobs:
           context: .
           file: docker/Dockerfile.build
           push: true
-          pull: true
+          build-args: CACHE_VERSION=${{env.CACHE_VERSION}}
           tags: ghcr.io/${{ github.repository }}/aurae-builder:latest,ghcr.io/${{ github.repository }}/aurae-builder:${{ steps.image-tag.outputs.tag}}
           cache-from: type=gha
           cache-to: type=gha,mode=max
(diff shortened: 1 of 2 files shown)


.github/workflows/build-with-containers.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           context: .
           file: docker/Dockerfile.build
           push: true
-          pull: true
```
jesse-peters, line 59: I removed `pull: true` on purpose here.

future-highway (approved):

PR #244, opened 2023-01-07
Absorb `Process` struct into `NestedAuraed`

future-highway:
Since `Executable` no longer uses `Process`, it did not make sense to keep it around as it was a place to share logic between `Executable` and `NestedAuraed`. This PR absorbs the parts that were used by `NestedAuraed` into it.

It also updates the CellSpec used for the tests so that no namespaces are un-shared, which allows the tests to not hang. The todo comment was updated to indicate we need further work on it.


auraed/src/runtime/cell_service/cells/mod.rs
```diff
@@ -59,7 +59,14 @@ impl CellSpec {
                 cpu_weight: CpuWeight::new(0),
                 cpuset_mems: CpusetMems::new("".into()),
             },
-            shared_namespaces: Default::default(), // nothing shared in default
+            shared_namespaces: SharedNamespaces {
```
dmah42, line 62: do you think we should still share nothing by default but share everything just for tests?

dmah42: or implement default for SharedNamespaces that does this (so we're still hiding it behind a "default".

dmah42 (approved):

auraed/src/runtime/cell_service/cells/mod.rs
future-highway: The default will still have everything as unshared, which I believe is what is wanted. 

This function is only for the tests, so that we don't need a new `CellSpec` each time. Having all shared namespaces allows the tests to not hang for now, which I think is a better position to be in because I was no longer running the tests prior to opening PRs.

I do think that we want to revisit this and see what the issue was, which ns was the cause, and is there a fix, which is why I updated the todo in the other file.

PR #246, opened 2023-01-07
Add go_package to v0 api proto

manuelcoppotelli:
In order to implement the gRPC go client, as per aurae-runtime/ae#13, it would be nice to add the `go_package` option to the proto files.
So that, the generated stubs are put in the correct go package.

### DIFF
--- a/api/v0/observe/observe.proto
+++ b/api/v0/observe/observe.proto
@@ -32,7 +32,7 @@ syntax = "proto3";
 
 package aurae.observe.v0;
 
-option go_package = "github.com/aurae-runtime/ae/client/pkg/api/observe/v0;observev0";
+option go_package = "github.com/aurae-runtime/ae/client/pkg/api/v0/observe;observev0";
 
 enum LogChannelType {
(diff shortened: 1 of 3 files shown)


krisnova (approved):

api/v0/observe/observe.proto
```diff
@@ -32,6 +32,8 @@ syntax = "proto3";
 
 package aurae.observe.v0;
 
+option go_package = "github.com/aurae-runtime/ae/client/pkg/api/observe/v0;observev0";
```
RRethy, line 35: We might want to be using a Go structure of `pkg/api/v0/observe` instead of `pkg/api/observe/v0` to match the proto structure. `observer|runtime|schedule` are all scoped under the same version.

manuelcoppotelli: Yes, it does make sense

krisnova: OH yes we sure mirror the same structure please

RRethy (approved):

krisnova (approved):

PR #247, opened 2023-01-08
add Health endpoint to discovery service

dmah42:
i think we might want to rename the schedule service to "discovery" or something similar as aurae itself won't be scheduling things (that will be a higher level service/client).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ stdlibdocs:
 	$(error "No /usr/local/bin/protoc-gen-doc, install from https://github.com/pseudomuto/protoc-gen-doc")
 else
 stdlibdocs:
-	protoc --plugin=/usr/local/bin/protoc-gen-doc -I api/v0/runtime -I api/v0/schedule -I api/v0/observe --doc_out=docs/stdlib/v0 --doc_opt=markdown,index.md:Ignore* api/v0/*/*.proto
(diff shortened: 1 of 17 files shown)


krisnova:
Strongly agree that we should rename. I think **Discovery** is a good name to start with. 

I also think we shouldn't use the term **Ping** unless we implement [ICMP RFC 792](https://www.rfc-editor.org/rfc/rfc792) and actually write a compliant "Echo" and "Echo Reply" implementation.

Maybe we call this **Announce** or **Alive** or **Heartbeat** (I am at a loss for good names at the moment but you get the idea).

dmah42:
> Strongly agree that we should rename. I think **Discovery** is a good name to start with.

ack

> 
> I also think we shouldn't use the term **Ping** unless we implement [ICMP RFC 792](https://www.rfc-editor.org/rfc/rfc792) and actually write a compliant "Echo" and "Echo Reply" implementation.

ack

> 
> Maybe we call this **Announce** or **Alive** or **Heartbeat** (I am at a loss for good names at the moment but you get the idea).

IsThereAnybodyOutThere

i think i will add it as "Health" and return both a 'healthy' boolean and the version, which will allow us to add a more complex health check down the road.

future-highway (approved):

PR #255, opened 2023-01-09
Introduce GracefulShutdown

future-highway:
Introducing the `GracefulShutdown` struct, responsible for cleaning up [insert what we need to clean up].

Also, we are now listening for both SIGTERM and SIGINT to trigger the graceful shutdown. #199 and the docs are not in agreement on the signal handling.

### DIFF
--- a/auraed/src/lib.rs
+++ b/auraed/src/lib.rs
@@ -81,11 +81,11 @@ use tonic::transport::{Certificate, Identity, Server, ServerTlsConfig};
 use tracing::{error, info, trace};
 
 mod discovery;
+mod graceful_shutdown;
 pub mod init;
 pub mod logging;
 mod observe;
 mod runtime;
-mod signal_handlers;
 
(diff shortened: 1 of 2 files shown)


auraed/src/lib.rs
```diff
@@ -230,6 +230,10 @@ impl AuraedRuntime {
         let discovery_service_server =
             DiscoveryServiceServer::new(discovery_service.clone());
 
+        let graceful_shutdown =
+            graceful_shutdown::GracefulShutdown::new(cell_service);
```
dmah42, line 234: should we just pass all services to this by default?

dmah42 (approved):

auraed/src/lib.rs
future-highway: That may be the end result, but we may get a nicer structure that ensures the right sequence is enforced on startup and shutdown.

Issue #254, opened 2023-01-09
We need to wait for the nested auraed to start listening on the socket


future-highway:
Trying to run cells_output.ts, surfaces an issue where the parent auraed is trying to connect to the child auraed before the child has a chance to setup and listen on the socket, resulting in a transport error.

future-highway:
#282

PR #256, opened 2023-01-09
Small miscellaneous changes

future-highway:
- Updated the example scripts to use cell names with the "ae-" prefix. This makes them easier to spot and remove manually if needed.
- Added an assert to help guard against changes in the clone crate's api (similar to std's command)
- ~~Added a temporary and crude workaround for #254 (sleep for 3 seconds)~~

### DIFF
--- a/auraed/src/runtime/cell_service/executables/auraed/nested_auraed.rs
+++ b/auraed/src/runtime/cell_service/executables/auraed/nested_auraed.rs
@@ -32,6 +32,7 @@ use crate::runtime::cell_service::executables::auraed::isolation_controls::{
     Isolation, IsolationControls,
 };
 use aurae_client::AuraeConfig;
+use clone3::Flags;
 use nix::{
     libc::SIGCHLD,
     sys::signal::{Signal, SIGKILL, SIGTERM},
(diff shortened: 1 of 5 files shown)


auraed/src/runtime/cell_service/executables/auraed/nested_auraed.rs
```diff
@@ -157,6 +161,10 @@ impl NestedAuraed {
                 let process = procfs::process::Process::new(pid)
                     .map_err(|e| io::Error::new(ErrorKind::Other, e))?;
 
+                // TODO: Remove this temporary solution to waiting for the nested auraed
+                //       to start listening on the socket
+                std::thread::sleep(Duration::from_secs(3));
```
krisnova, line 166: This is very scary to start bringing into a runtime daemon, even if its only temporary. 

Maybe we should remove this in this PR and bring in another PR where we use the discovery endpoint?

future-highway: Sure

dmah42 (approved):

Issue #216, opened 2022-12-30
Consider CRI


krisnova:
Consider a CRI shim of sorts that sits on top of Aurae in the future? https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto

krisnova:
Closed with #308

PR #264, opened 2023-01-11
Introduce CRI (Kubernetes 1.26) Container Runtime Interface to Aurae

krisnova:
This pull request will use the new static musl build to re-introduce the `/pods` directory back into the `auraed`.

We will need to keep the build passing in this PR so we have a static binary when we are finished.


krisnova:
My plan for this:

 - Adopt CRI for our pod service https://github.com/aurae-runtime/aurae/issues/216
 - Use the new `spawn` module in favor of a traditional "Pause Container" #269 
 - Move `/pods` directory -> `auraed/src/runtime/pod_service/pods` where it started
 - Figure out how to compile libcontainer with musl (need to remove libseccomp and libdbus https://github.com/containers/youki/pull/1472)

PR #271, opened 2023-01-12
Updates makefile with aarch64 targets

future-highway:
- Created makefile commands for `--target aarch64-unknown-linux-musl` where needed
- Updated `make test` to use `--target x86_64-unknown-linux-musl` when not testing `auraescript`
- Removed `make clean-auraescript` and `make clean-auraed` (they don't do anything; left over from when we had multiple repos)
- Reactivated the tests and commented for when they are ignored
- Updated M1 development guide

---

Multiple tests that are ignored in CI are failing, in relation to cgroups-rs. An example output (others are the same):

>---- runtime::cell_service::cells::cells::tests::test_allocate stdout ----
thread 'runtime::cell_service::cells::cells::tests::test_allocate' panicked at 'inner from builder: Error { kind: WriteFailed("cpu.weight", "0"), cause: Some(Os { code: 34, kind: Uncategorized, message: "Result not representable" }) }', auraed/src/runtime/cell_service/cells/cgroups/cgroup.rs:64:29

Seems like we need to update our validation. (edit: may just be the tests, as they skip validation)


Makefile
```diff
@@ -184,3 +179,23 @@ container: ## Build the container defined in hack/container
...
+
+build-m: musl-m auraed-m auraescript ## Build and install (debug) (+musl) 🎉 (aarch64-unknown-linux-musl)
+prcheck-m: build-m lint test-all-m ## Meant to mimic the GHA checks that are performed when opening a PR (aarch64-unknown-linux-musl)
+
+musl-m: ## Add target for musl (aarch64-unknown-linux-musl)
+	rustup target add aarch64-unknown-linux-musl
+
+auraed-m: proto ## Initialize and static-compile auraed with musl (aarch64-unknown-linux-musl)
+	@$(cargo) clippy -p auraed
+	@$(cargo) install --target aarch64-unknown-linux-musl --path ./auraed --debug --force
```
dmah42, line 193: can we not do:

```
uname_m := $(shell uname -m)

auraed: proto
  @$(cargo) install --target $(uname_m)-unknown-linux-musl --path ./auraed --debug --force
```

?
--target 
```

dmah42 (approved):

PR #272, opened 2023-01-13
Cells & Cgroups

future-highway:
Currently cgroups are not working. It seems like cgroup-rs 0.2 was silently failing. 0.3 no longer silently fails, and we were getting an error when setting `cpu.weight = 0`. That bug has been fixed, but now we are hitting another error:

>Error: Error: status: Internal, message: "cell 'ae-sleeper-cell' allocation was aborted: unable to write to a control group file cgroup.threads, value 200680 caused by: Os { code: 95, kind: Uncategorized, message: \"Not supported\" }", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Fri, 13 Jan 2023 03:28:01 GMT", "content-length": "0"} }

After this error is returned:
- ~~The cgroup exits on the system in an unknown configuration~~ (we now successfully remove the cgroup if adding the process fails)
- ~~Nested auraed is started~~ (nested auraed is killed if on `Err`)
- ~~Nested auraed is not in the cgroup~~ (nested auraed is either in the cgroup or killed)

---

Other issues fixed:
- `Uri::try_parse` was succeeding on the unix socket path, and then erroring when trying to create a channel from a Uri
- `make start` used `sudo` instead of `sudo -E` which resulted in not being able to find the config file

---

In trying to fix the cgroup errors, I modeled out the api messages as suggested in #228. Naming of fields were updated to reflect v2 cgroup wording.


krisnova (dismissed):

auraed/src/runtime/cell_service/cells/cgroups/cgroup.rs
```diff
@@ -42,26 +46,57 @@ pub struct Cgroup {
...
+            let builder = builder.cpu();
+
+            let builder = if let Some(weight) = weight {
+                builder.shares(weight.into_inner())
+            } else {
+                builder
+            };
+
+            let builder = if let Some(max) = max {
+                builder.quota(max.into_inner()).period(1_000_000) // microseconds in a second
```
dmah42, line 68: note i have a PR creating a const for this with docs.  feel free to absorb it or merge it.

auraed/src/runtime/cell_service/cells/cgroups/max.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl ValidatedField<i64> for Max {
+    fn validate(
+        input: Option<i64>,
+        field_name: &str,
+        parent_name: Option<&str>,
+    ) -> Result<Self, ValidationError> {
+        let input = validation::required(input, field_name, parent_name)?;
+
+        validation::minimum_value(input, 0, "units", field_name, parent_name)?;
```
dmah42, line 57: maybe i've misread this but shouldn't this _also_ have a validation for maximum_value 10000?

auraed/src/runtime/cell_service/cells/cgroups/cgroup.rs
future-highway: already merged and rebasing on it

auraed/src/runtime/cell_service/cells/cgroups/max.rs
future-highway: There doesn't seem to be a requirement for that:

>Limits are in the range [0, max] and defaults to “max”, which is noop.
>
>As limits can be over-committed, all configuration combinations are valid and there is no reason to reject configuration changes or process migrations.

But, it doesn't seem to make sense to go over the maximum allowed weight (10000).

dmah42: that's what i was thinking. but maybe there are other "max"es that are not 10k.

dmah42 (approved):

auraed/src/runtime/cell_service/cells/cgroups/max.rs
future-highway: Hmm, reading the [docs](https://docs.kernel.org/admin-guide/cgroup-v2.html#resource-distribution-models) again. "max" is used in multiple contexts. Here we are referring to limits, so I think we should leave the validation, but rename the field/struct to limit

api/v0/runtime/runtime.proto
```diff
@@ -299,7 +299,7 @@ message CpuController {
   // In one period (1_000_000), how much can the tasks run.
   //
   // * Minimum: 0
-  optional int64 max = 2;
+  optional int64 limit = 2;
```
dmah42, line 302: i know you decided to rename this, but if the point is to stay close to the cgroups v2 API then this should absolutely be `max`.  but it also should take the value "max", not just an integer.

future-highway: Good point re naming.

For accepted values, "max" is the default so equivalent of not settings limit. Thoughts?

dmah42: you can also set it explicitly to "max".  i think it's ok to document that as the default as i can't see a user wanting to explicitly set it to unlimited...

future-highway: I'll improve the documentation. We could later introduce -1 as asking to explicitly set max, without breaking the api, but that wouldn't match the underlying api.

future-highway: Another option would be to later introduce a oneof, but I understand we are trying to stay away from oneofs.

Unless we introduce the ability to adjust the resources of a cell, I don't think an explicit max will ever be needed and even then, we can change the implementation to always set "max" when the field is `None` instead of relying on the default

dmah42 (approved):

PR #274, opened 2023-01-13
Grpc health service

future-highway:
Implement the [GRPC Health Checking Protocol](https://github.com/grpc/grpc/blob/master/doc/health-checking.md)

- `aurae-client` supports both `check` and `watch`
- `auraescript` only supports `check` as I haven't looked into how/if deno handles streaming


auraed/src/lib.rs
```diff
@@ -237,25 +237,32 @@ impl AuraedRuntime {
...
+        health_reporter.set_serving::<CellServiceServer<CellService>>().await;
 
         let discovery_service = DiscoveryService::new();
         let discovery_service_server =
             DiscoveryServiceServer::new(discovery_service.clone());
+        health_reporter.set_serving::<DiscoveryServiceServer<DiscoveryService>>().await;
 
         let pod_service = PodService::new(self.runtime_dir.clone());
         let pod_service_server = PodServiceServer::new(pod_service.clone());
+        health_reporter.set_serving::<PodServiceServer<PodService>>().await;
```
dmah42, line 254: should this come after we've spawned the server below?  it likely doesn't matter but...

dmah42 (approved):

auraed/src/lib.rs
future-highway: I get that, but you can't actually read from the health reporter, so I think it's okay to leave it here and keep each services setup grouped.

PR #278, opened 2023-01-14
Remove 005, fix 007 container path reference

jesse-peters:
- Fixes GHA concurrency issue with docker container builds
- Adds back readme images
- Some small naming issues on the GHA steps


krisnova:
Please squash into a single commit once you are finished

jesse-peters:
Thanks, that's a bad habit I'm still getting past. I'll keep working on it!

jesse-peters:
Looking at why 036 wasnt run

jesse-peters:
Alright we should be set. Whats best practice reagarding squashing in a case like this where there is a single commit after fixing a merge issue?

.github/workflows/007-cargo-udeps-aurae-builder-make-build-check-deps.yml
```diff
@@ -41,9 +41,9 @@ env:
  CARGO_TERM_COLOR: always
 jobs:
  build-container:
-   uses: ./.github/workflows/031-ubuntu-aurae-builder-image-ubuntu-aurae-builder-docker-cargo-udeps-build.yml
- build-docs-with-container:
-   name: Build Docs
+   uses: ./.github/workflows/031-ubuntu-aurae-builder-image-ubuntu-docker-cargo-udeps-build.yml
+ check-udeps:
+   name: Check udeps
```
future-highway, line 46: *I can only comment on lines with a change*

Can you remove the build step (ln 69-70). cargo-udeps does its own build and it won't share the build from `make build` so we are building an extra time without benefit.

future-highway (changes requested):
Blocking to indicate that this should not be merged until the PR check based on workflow 035 is set as required by an admin, replacing 005, which this PR removes.

jesse-peters:
Thanks @future-highway

.github/workflows/007-cargo-udeps-aurae-builder-make-build-check-deps.yml
jesse-peters: I can, but my concern with that is that cargo-udeps needs to install `libssl-dev`, and I didn't want to pollute the `make build` cache. I may have misunderstood the directions to not add those, so please correct me if I'm wrong 

https://github.com/aurae-runtime/aurae/blob/main/images/Dockerfile.build#L110

krisnova: Its fine for things like cargo-udeps I guess. I want to make sure we aren't using GNU libraries for OUR dependencies.

krisnova (approved):

.github/workflows/007-cargo-udeps-aurae-builder-make-build-check-deps.yml
jesse-peters: It was a misunderstanding on my part, we resolved it in discord. We'll continue to split the docker image into stages so the build stage won't have the extra libraries. I was using the incorrect make command in the GHA job which resolved our issue.

PR #287, opened 2023-01-16
aer

### LINKED ISSUE #266: Create "aer"
Can we generate a command line client for Aurae using the `aurae-client` module?

Should we rename the module to `client-rust` to match the rest of the Aurae project landscape?

future-highway:
Unfortunately I'm not seeing a path for generating the cli via macros. Doing it manually is not hard, just not ideal.

There are some opinions here in terms of aliases and short flags, but I don't know if they are the right decisions.

Example commands:
```bash
aer cell allocate ae-test-cell --isolate-process --isolate-network
aer cell start ae-test-cell ps-aux -c "ls /proc"
aer cell stop ae-test-cell ps-aux
aer cell free ae-test-cell
```

Example server streaming command:
```bash
aer health watch
```

closes #266


aer/src/runtime/cell_service.rs
```diff
@@ -0,0 +1,151 @@
...
+    Cell, CellServiceAllocateRequest, CellServiceFreeRequest,
+    CellServiceStartRequest, CellServiceStopRequest, CpuController,
+    CpusetController, Executable,
+};
+use clap::Subcommand;
+
+// We'd like to macro this struct and the impl from the CellService in the proto.
+// I'm (future-highway) not sure how, even though this seems almost entirely boilerplate.
+//
+// Maybe macros is the wrong approach.
```
krisnova, line 43: Hrmmm I see what you mean here. Having to define the `CellServiceCommands` enum for each command seems like a lot of work. This feels similar to what I originally was trying to do with AuraeScript before it become much more beautiful like it is now.

Why won't macros work exactly? This enum looks like something we should be able to generate with a macro?

future-highway: For proc_macros, which are the macros that are in their own crates (as opposed to `macro_rules`) we have 3 choices of macros. Two use attributes (e.g., `#[derive(MyMacro)]` and `#[my_macro]`), and need a struct or enum for us to "attach" them. We can overcome this hurdle by generating the rust code using build.rs like we did prior to buf, but then we would be left with attaching attributes in the build.rs, which is also not great.

The other option is to make a macro like those used in `auraescript` and `aurae-client` where we essentially define our own syntax. That may make the code shorter, but not necessarily better to deal with (we'd need to update the macro each time we have a feature of `clap` that isn't made available in the macro syntax yet).

If we had a proto parser, then we may be able to significantly reduce the code needed and have a better dev experience, but I was unable to find a rust crate that seemed suitable.

future-highway: Well, now we have a proto parser in use for aurae-client and soon auraescript.

I think the creation of macro of aer to use it will be more work then the conversion of the other crates, but it should be technically possible. It won't be as minimal as we still want to be able to use clap attributes I would think, but it should be possible to significantly reduce the code needed.

future-highway:
I think we should merge this as is (without a macro) if we want `aer` regardless of whether it is autogenerated.

See #298 for more context.

dmah42 (approved):

PR #292, opened 2023-01-17
Generate auraescript from protos

future-highway:
The auraescript macro was relying on convention to locate the generated typescript files. As we brought in 3rd party protos, they do not match the project's conventions and thus needed carve outs.

This PR makes the path explicit in the macro instead of relying on conventions.

### DIFF
--- a/auraescript/src/cri.rs
+++ b/auraescript/src/cri.rs
@@ -30,7 +30,7 @@
 
 // TODO: macro doesn't support streaming. Does deno?
 macros::ops_generator!(
-    kubernetes::cri,
+    kubernetes::cri("kubernetes/cri/v1/release-1.26.ts"),
     {
         RuntimeService,
         version(VersionRequest) -> VersionResponse,
--- a/auraescript/src/discovery.rs
+++ b/auraescript/src/discovery.rs
@@ -29,9 +29,7 @@
 \* -------------------------------------------------------------------------- */
 
 macros::ops_generator!(
-    discovery("v0/discovery/discovery.ts"),
-    {
-        DiscoveryService,
-        discover(DiscoverRequest) -> DiscoverResponse,
-    },
+    "../api/v0/discovery/discovery.proto",
+    discovery,
+    DiscoveryService,
 );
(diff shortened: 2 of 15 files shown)


auraescript/src/cri.rs
```diff
@@ -30,7 +30,7 @@
 
 // TODO: macro doesn't support streaming. Does deno?
 macros::ops_generator!(
-    kubernetes::cri,
+    kubernetes::cri("gen/kubernetes/cri/v1/release-1.26.ts"),
```
dmah42, line 33: these all start with gen and you add "gen" to the out_dir above. do we get double-gen?

auraescript/src/discovery.rs
```diff
@@ -29,7 +29,7 @@
 \* -------------------------------------------------------------------------- */
 
 macros::ops_generator!(
-    discovery,
+    discovery("gen/v0/discovery/discovery.ts"),
```
dmah42, line 32: it is a shame we can't generate this from the proto (version + service-name).  maybe the next iteration is a pointer to the proto and then a macro that reads the proto and parses out the services and version :D

auraescript/src/cri.rs
future-highway: We don't get a double "gen" as I took that out of the macro logic for locating the ts. I dropped it in hopes that I could have docs with examples, which I got to, but then realized the macro outputs files, and the example would also output files when testing as cargo tests rust code in docs too, so I dropped the docs example.

In other words, I can put the assumption back into the macro if that is preferred?

dmah42: i think it's a bit easier than trying to remember to put "gen" in front every time. but it's a small thing.

auraescript/src/discovery.rs
future-highway: Ran into this with the aer PR...I wasn't able to find a suitable rust crate that parses proto files. Everything seems to rely on protoc.

future-highway: As I'm thinking about it...we don't really need a good proto parser, just something to read the proto file and extract what we want. It may be possible with the nom crate (or another one whose name I'm forgetting at the moment but is supposed to be easier to use).

dmah42 (approved):

future-highway:
I'm trying to update this to also read from the proto files

dmah42 (approved):

PR #297, opened 2023-01-18
replace 'health' with 'discover'

dmah42:


api/v0/discovery/discovery.proto
```diff
@@ -1,5 +1,5 @@
 /* -------------------------------------------------------------------------- *\
- *        Apache 2.0 License Copyright © 2022-2023 The Aurae Authors          *
+ *             Apache 2.0 License Copyright © 2022 The Aurae Authors          *
```
future-highway, line 2: I know we have the upcoming PR to remove dates, but do we want to leave it with the range for now.

future-highway (approved):

api/v0/discovery/discovery.proto
dmah42: this was a mistake. reverted.

future-highway (approved):

PR #302, opened 2023-01-20
Allow multiple cell nesting 

future-highway:
We are having issues with nesting cells at a depth greater than 1. This PR is meant to fix that.

The issue stems from the "no internal processes" rule for v2 [cgroups](https://man7.org/linux/man-pages/man7/cgroups.7.html). 
> ...with the exception of the root cgroup, processes may reside only in leaf nodes (cgroups that do not themselves contain child cgroups).

Essentially, our nested auraeds, themselves a process, are not allowed to create cgroups. With this PR, only the root auraed will be responsible for creating cgroups, and the nested auraed will be responsible for spawning executables.

I'm want to use this PR to also fix graceful shutdown. We haven't been shutting down the spawned executables, which is causing issues with cleaning up the cgroups.

---

edit: `GracefulShutdown` now broadcasts stop to all executables (which sends a SIGKILL), allowing the cgroup. to be cleaned up.

---

NOTE: The api for `CellService`'s start and stop has been updated, making `cell_name` optional. A null `cell_name` means auraed should spawn the process in its own cgroup. For the auraed that is true pid 1, that means the executable is not in a cell (it is in the root cgroup).

### DIFF
--- a/aer/src/runtime/cell_service.rs
+++ b/aer/src/runtime/cell_service.rs
@@ -129,7 +129,7 @@ impl CellServiceCommands {
                 executable_description,
             } => {
                 let req = CellServiceStartRequest {
-                    cell_name,
+                    cell_name: Some(cell_name),
                     executable: Some(Executable {
                         name: executable_name,
                         command: executable_command,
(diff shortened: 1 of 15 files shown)


krisnova (approved):
LGTM

PR #303, opened 2023-01-21
Add git cli to dockerfile. Adds release GHA job

jesse-peters:
This is related to the release project [PR](https://github.com/aurae-runtime/release/pull/1) that sets up the main release cycle. 

It will work independently of that though if a release is created here manually.


Makefile
```diff
@@ -205,4 +205,15 @@ check-deps: ## Check if there are any unused dependencies in Cargo.toml
...
 test-workflow: ## Tests a github actions workflow locally using `act`
-	@act -W ./.github/workflows/$(file)
\ No newline at end of file
+	@act -W ./.github/workflows/$(file)
+
+.PHONY: stage-release-artifacts
+stage-release-artifacts: ## Stage release artifacts
+	@mkdir -p /tmp/release
+	@cp ~/.cargo/bin/auraed /tmp/release/auraed-$(tag)-$(uname_m)-unknown-linux-musl
+	@cp ~/.cargo/bin/auraescript /tmp/release/auraescript-$(tag)-$(uname_m)-unknown-linux-musl
```
future-highway, line 214: Does auraescript compile with musl for you? `cargo build -p auraescript --target aarch64-unknown-linux-musl` fails to compile for me.

jesse-peters: No, looks like I got an error too. I ended up grabbing a used Xeon workstation to throw a dev environment on, so I've been using that more.

jesse-peters:
Accidentally locked. Disregard

Makefile
future-highway: Does it work on x86? I thought it wouldn't work at all with musl

jesse-peters: Yea it seems to have. I ran 'make release' and had an auraescript binary that worked as an output.

future-highway: Just checked, `make release` doesn't use musl for auraescript (it uses the default which will be gnu)

future-highway (approved):

PR #308, opened 2023-01-22
RunPodSandbox CRI Implementation (Starting point)

krisnova:
This pull request will begin to plumb container scheduling through Aurae using the Kubernetes container runtime interface (CRI). 


krisnova:
This is mostly a refactor, and scratches the surface of the work we need to do with `RunPodSandbox` however it gets us to the point where the big pieces are all plumbed through and we can schedule a "pod" with Aurae. 

Now we can iterate on the various sub-sections of work (EG: linux config, hostname, dns, etc) independently and in parallel.

PR #311, opened 2023-01-23
Fix make steps for artifacts

jesse-peters:


.github/workflows/050-release-ubuntu-make-build-release.yml
```diff
@@ -65,7 +62,7 @@ jobs:
                 key: release-${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
             - name: Make release artifacts
-              run: make release
+              run: make musl build release
```
future-highway, line 65: 'make musl proto release' should work. Better yet, add the proto step to the makefile like it is for other make commands

jesse-peters: Ah right, I forgot the proto step is better than the build. Thanks, Ill fix that.

future-highway (approved):

PR #318, opened 2023-01-24
`make proto-lint` clean

dmah42:


api/grpc/health/v1/health.proto
```diff
@@ -25,24 +25,34 @@ option java_multiple_files = true;
...
-    NOT_SERVING = 2;
-    SERVICE_UNKNOWN = 3;  // Used only by the Watch method.
-  }
+message WatchResponse {
   ServingStatus status = 1;
 }
 
-service Health {
+
+service HealthService {
```
future-highway, line 52: I think this is a breaking change. Overall, I don't think we can or should change this file as it is not ours, but a shared API that any gRPC health client will rely on. 

Also, auraed doesn't actually use this file to create the endpoint; tonic-health has its own copy embedded into the create. This file is just for the macros.

future-highway (changes requested):

api/grpc/health/v1/health.proto
dmah42: reverted

future-highway (approved):

PR #319, opened 2023-01-24
Refactor Makefile

future-highway:
An attempt at organizing the Makefile.


.github/workflows/035-cargo-test-lint-aurae-builder-make-musl-proto-test-lint.yml
```diff
@@ -30,43 +30,43 @@
...
-       # The goal is for this to be easy to find from the GitHub dashboard.
-       # Instead of adding more commands here, consider another make target or a new YAML file
-       # named with a good name.
-       run: make musl proto test lint
+      - name: Cargo Test Lint [make test lint]
+        # This should remain the only command we execute as this matches the title of the file.
+        # The goal is for this to be easy to find from the GitHub dashboard.
+        # Instead of adding more commands here, consider another make target or a new YAML file
+        # named with a good name.
+        run: make musl lint test
```
dmah42, line 72: the file should be renamed if we're dropping proto.  i do wonder why this isn't just "make test" and assume the dependent jobs are correctly run.  i think this would be safer, no?

Makefile
```diff
@@ -40,84 +40,57 @@ cri_version   =  release-1.26
...
-	@$(cargo) test -p auraescript
+# Setup Commands
 
-test-all: musl proto ## Run the tests (including ignored)
-	@sudo -E $(cargo) test --target $(uname_m)-unknown-linux-musl --workspace --exclude auraescript -- --include-ignored
-	@$(cargo) test -p auraescript -- --include-ignored
+.PHONY: pki
+pki: certs ## Alias for certs
+.PHONY: certs
+certs: clean-certs ## Generate x509 mTLS certs in /pki directory
```
dmah42, line 88: do we need to clean before creation?  i'm always a bit suspicious of targets that auto-clean.

Makefile
```diff
@@ -127,69 +100,187 @@ config: ## Set up default config
...
 .PHONY: proto-lint
 proto-lint: ## Lint protobuf schemas
 	buf lint api
 
-.PHONY: clean
-clean: clean-certs
-	@rm -rf aurae-proto/src/gen/*
-	@cargo clean
+.PHONY: proto-vendor-cri
+proto-vendor-cri: ## Copy the CRI interface from upstream
```
dmah42, line 138: should this be `proto-vendor` and include the grpc proto?

.github/workflows/035-cargo-test-lint-aurae-builder-make-musl-proto-test-lint.yml
future-highway: I like that idea. I'll add lint to test

Makefile
```diff
@@ -127,69 +100,187 @@ config: ## Set up default config
...
-.PHONY: help
-help:  ## Show help messages for make targets
-	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(firstword $(MAKEFILE_LIST)) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[32m%-30s\033[0m %s\n", $$1, $$2}'
+.PHONY: auraed-release
+auraed-release:
+	@$(cargo) install --target $(uname_m)-unknown-linux-musl --path ./auraed --force
+
+#------------------------------------------------------------------------------#
+
+# AuraeScript Commands
```
dmah42, line 170: maybe bring back the auraescript/Makefile and then:

```
TARGETS := all clean lint test test-all debug release
SUBDIRS := $(wildcard */.)

$(TARGETS): $(SUBDIRS)
$(SUBDIRS):
        $(MAKE) -C $@ $(MAKECMDGOALS)

.PHONY: $(TARGETS) $(SUBDIRS)
```

this would allow "make test" to iterate through all the subdirectories calling "make test" in each.

Makefile
future-highway: Maybe. I'll look into vendoring the grpc health proto

.github/workflows/001-cargo-install-ubuntu-make-build.yml
```diff
@@ -98,4 +98,4 @@ jobs:
         # The goal is for this to be easy to find from the GitHub dashboard.
         # Instead of adding more commands here, consider another make target or a new YAML file
         # named with a good name.
-        run: make build
+        run: make musl build
```
dmah42, line 101: if build needs musl, build should depend on musl :)

Makefile
```diff
@@ -40,84 +40,57 @@ cri_version   =  release-1.26
...
+docs: proto docs-crates docs-stdlib docs-other docs-lint ## Assemble all the /docs for the website locally.
 
-crate: musl ## Build the crate (documentation)
-	$(cargo) doc --target $(uname_m)-unknown-linux-musl --no-deps --package auraed
-	$(cargo) doc --no-deps --package auraescript
-	$(cargo) doc --no-deps --package aurae-client
-	$(cargo) doc --no-deps --package aer
-	cp -rv target/doc/* docs/crate
+.PHONY: prcheck
+prcheck: install lint test-all docs ## Meant to mimic the GHA checks (includes ignored tests)
```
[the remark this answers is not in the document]
future-highway: I'm not sure why we've gotten as minimal per check as we have, but `install` is its own check to restrict the dependencies that are available to ensure we can statically compile `auraed`.

Makefile
future-highway: I'm not sure I 100% understand this, so I'll ask that we put it off for another PR if we want it.

dmah42: sorry, i got a bit carried away.  that's fine, i can do this later.  it's an optimization to make this all a bit cleaner.

.github/workflows/001-cargo-install-ubuntu-make-build.yml
future-highway: You are right, but I hate that we are calling it every time locally (even thought it's cheap). I'll update

.github/workflows/035-cargo-test-aurae-builder-make-musl-test.yml
```diff
@@ -28,45 +28,45 @@
...
-       # The goal is for this to be easy to find from the GitHub dashboard.
-       # Instead of adding more commands here, consider another make target or a new YAML file
-       # named with a good name.
-       run: make musl proto test lint
+      - name: Cargo Test [make test]
+        # This should remain the only command we execute as this matches the title of the file.
+        # The goal is for this to be easy to find from the GitHub dashboard.
+        # Instead of adding more commands here, consider another make target or a new YAML file
+        # named with a good name.
+        run: make test
```
dmah42, line 72: another rename necessary? just "...make-test.yaml" now.

Makefile
future-highway: I assume `certs` will fail (the clean step was there before) if the pki directory is not empty. I don't know what the correct approach should be and don't have an opinion. Your call.

Makefile
```diff
@@ -40,84 +40,57 @@ cri_version   =  release-1.26
...
 
 .PHONY: all
 all: install ## Build and install (debug) 🎉
 
-.PHONY: install
-install: build ## Build and install (debug) 🎉
-
 .PHONY: build
-build: musl auraed auraescript ## Build and install (debug) (+musl) 🎉
+build: install ## Build and install (debug) (+musl) 🎉
```
dmah42, line 51: is this just an alias for `install`?  i would expect install to move the binaries to /usr/local/bin or something .. do we need both?

Makefile
```diff
@@ -40,84 +40,57 @@ cri_version   =  release-1.26
...
-musl: ## Add target for musl
-	rustup target add $(uname_m)-unknown-linux-musl
+.PHONY: test-all
+test-all: lint libs-test-all auraed-test-all auraescript-test-all aer-test-all ## Run lints and tests (includes ignored tests)
 
-.PHONY: auraed
-auraed: musl proto ## Initialize and static-compile auraed with musl
-	@$(cargo) clippy --target $(uname_m)-unknown-linux-musl -p auraed
-	@$(cargo) install --target $(uname_m)-unknown-linux-musl --path ./auraed --debug --force
+.PHONY: install
```
dmah42, line 69: why is `install` debug but `release` release?  why does `install` not lint and test before building?

Makefile
```diff
@@ -40,84 +40,57 @@ cri_version   =  release-1.26
...
-	$(cargo) install --path ./auraescript
+.PHONY: clean
+clean: clean-certs clean-gens clean-crates ## Clean the repo
 
-.PHONY: auraescript
-auraescript: proto ## Initialize and compile auraescript
-	@$(cargo) clippy -p auraescript
-	@$(cargo) install --path ./auraescript --debug --force
+.PHONY: lint
+lint: musl proto proto-lint libs-lint auraed-lint auraescript-lint aer-lint ## Run all lints
```
dmah42, line 61: lint shouldn't need to depend on `proto` should it?

Makefile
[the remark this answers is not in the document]
dmah42: i don't know what this means, sorry.

if we don't use `prcheck` anywhere can we remove it?

Makefile
```diff
@@ -127,69 +100,194 @@ config: ## Set up default config
...
+libs-test-all:
+	@$(cargo) test --workspace --exclude auraed --exclude auraescript --exclude aer -- --include-ignored
+
+#------------------------------------------------------------------------------#
+
+# Commands for testing
+
+.PHONY: test-workflow
+test-workflow: ## Tests a github actions workflow locally using `act`
+	@act -W ./.github/workflows/$(file)
```
dmah42, line 251: where does `file` come from?  and do we need this to be a `make` command? can't i just run this myself if i need to test a workflow?

Makefile
```diff
@@ -215,25 +313,67 @@ oci-image-build-raw: ## Plain Jane oci build
 container: ## Build the container defined in hack/container
 	./hack/container
 
-.PHONY: check-deps
-check-deps: musl ## Check if there are any unused dependencies in Cargo.toml
-#	cargo +nightly udeps --target $(uname_m)-unknown-linux-musl --package auraed
-#	cargo +nightly udeps --package auraescript
-#	cargo +nightly udeps --package aurae-client
-#	cargo +nightly udeps --package aer
+.PHONY: spawn
```
dmah42, line 316: if nothing depends on these targets, can we drop them to keep the Makefile a bit cleaner?

Makefile
```diff
@@ -215,25 +313,67 @@ oci-image-build-raw: ## Plain Jane oci build
...
+.PHONY: headers-check
+headers-check: ## Only check for problematic files.
+	./hack/headers-check
+
+.PHONY: headers-write
+headers-write: ## Fix any problematic files blindly.
+	./hack/headers-write
+
+.PHONY: start
+start:
```
dmah42, line 367: auraed-start to match the other targets, i think?

Makefile
future-highway: I don't think `proto-lint` needs `proto`, but I'm pretty sure the lints for auraed, auraescript, & aer, require up-to-date proto files.

Makefile
future-highway: These are being used by Nóva for developing the cri services.

Makefile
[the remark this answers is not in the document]
future-highway: I use it regularly, especially before pushing as I can run the checks locally much more quickly than waiting for the PR checks 🙂.

Makefile
dmah42: i resolved this but then i realised: if they do then we should capture that in {auraed,auraescript,aer}-lint targets as dependencies.  make will only do it once (i think).

future-highway: I'll test

future-highway: It is true, and `proto` was already calling `proto-lint` which I realize I put in because I think the lint feedback should be immediate on regenerating the proto files

Makefile
future-highway: The missing lint was an error in my refactoring. Should be fixed now.

Why it doesn't test boils down to not breaking peoples workflow. We used to have a `default` which was an alias for `all` which was an alias for `install`, so running `make` would lint and install auraed and auraescript (I have added aer now too), but not test.

We no longer have the default, so maybe my attempt at not effecting anyone's habits is moot?

Makefile
future-highway: Yes, it is just an alias. `install` installs the binaries at, for example, `$(HOME)/.cargo/bin/auraed`.

I'll update build to not run the tests, and have install run tests. I think that's a good middle ground.

Makefile
future-highway: 🤷

Makefile
future-highway: I updated `build` to not include tests (only lints) and `install` now has all three (lints, tests and install)

dmah42 (dismissed):

Makefile
jesse-peters: I added this, the idea was to run github actions tests using `make test-workflow file=001-cargo-install-ubuntu-make-install.yml` to test an individual file. If its too clunky it can be removed.

krisnova (approved):

Issue #217, opened 2022-12-30
Add "list" method


krisnova:
We need a way to query the cache and list all known pods, containers, cells, and executables known to a specific auraed. 

If list works recursively we should be able to build a graph if we query the "root" auraed.

 - Introduce list method to runtime.proto and return a set of responses

krisnova:
This is becoming more relevant.

We need to be able to "list" cells. See #308 where we introduce list pods style methods with the CRI. We should be able to mirror this functionality with Aurae cells and executables.

JeroenSoeters:
@krisnova should we close this issue?

PR #322, opened 2023-01-27
Introduce "list" api for cells

JeroenSoeters:
This PR is introducing a "list" method to the API (https://github.com/aurae-runtime/aurae/issues/217). As of now this API does not yet work recursively, as in, only the cells of the main `auraed` process are reported. Would be good to get some feedback on the API and this code in general.


auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -237,6 +241,29 @@ impl CellService {
...
+    #[tracing::instrument(skip(self))]
+    async fn list(&self) -> Result<CellServiceListResponse> {
+        let cells = self.cells.lock().await;
+
+        
+        Ok(CellServiceListResponse { 
+            cells: cells.entries()
+                        .iter()
+                        .map(|cell| 
+                            Cell {
```
dmah42, line 254: there might be a very good reason for this: why do we need this map at all? or why is in not

```
.map(|cell| cell.clone())

File: auraed/src/cells/cell_service/cells/cells_cache.rs [hunk 2]
```diff
@@ -63,4 +63,6 @@ pub trait CellsCache {
 
     /// Sends a [SIGKILL] to all Cells, ignoring any errors.
     fn broadcast_kill(&mut self);
+
+    fn entries(&self) -> Vec<&Cell>;
```
future-highway, line 67: `entries` makes it seems like the [Entry api](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html) will be exposed. I'd call it `values`.

auraed/src/cells/cell_service/cells/cells.rs
```diff
@@ -243,6 +244,10 @@ impl CellsCache for Cells {
     fn broadcast_kill(&mut self) {
         self.broadcast_kill()
     }
+
+    fn entries(&self) -> Vec<&Cell> {
+        self.cache.values().by_ref().collect_vec()
```
future-highway, line 249:
```suggestion
        self.cache.values().collect()
```

auraed/src/cells/cell_service/cells/cell.rs
```diff
@@ -229,6 +233,11 @@ impl CellsCache for Cell {
 
         children.broadcast_kill()
     }
+
+    fn entries(&self) -> Vec<&Cell> {
+        //self.cache.iter().collect_vec()
+        todo!()
```
future-highway, line 239: No need to actually do work here. Follow the pattern you see in the other functions. Check if we are in the allocated state, if so, call the function on children.

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -237,6 +241,29 @@ impl CellService {
...
         Ok(())
     }
+
+    #[tracing::instrument(skip(self))]
+    async fn list(&self) -> Result<CellServiceListResponse> {
+        let cells = self.cells.lock().await;
+
+        
+        Ok(CellServiceListResponse { 
+            cells: cells.entries()
```
future-highway, line 251: This block could benefit from being formatted with rustfmt. If you are using VSCode you should be able to set it to run on save (https://stackoverflow.com/questions/67859926/how-to-run-cargo-fmt-on-save-in-vscode)

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -237,6 +241,29 @@ impl CellService {
...
+                        .iter()
+                        .map(|cell| 
+                            Cell { 
+                                name: cell.name().to_string(), 
+                                cpu: Some(CpuController { 
+                                    weight: cell.spec().cgroup_spec.cpu.as_ref().and_then(|cpu| cpu.weight.as_ref().map(|w| w.into_inner())),
```
future-highway, line 257: Instead of having `cell.spec().cgroup_spec.cpu.as_ref()` multiple times, I suggest making a variable to hold the reference.

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -237,6 +241,29 @@ impl CellService {
...
+                                name: cell.name().to_string(), 
+                                cpu: Some(CpuController { 
+                                    weight: cell.spec().cgroup_spec.cpu.as_ref().and_then(|cpu| cpu.weight.as_ref().map(|w| w.into_inner())), 
+                                    max: cell.spec().cgroup_spec.cpu.as_ref().and_then(|cpu| cpu.max.as_ref().map(|m| m.into_inner())) }), 
+                                cpuset: Some(CpusetController { cpus: Some(String::from("")), mems: Some(String::from("")) }), 
+                                isolate_process: cell.spec().iso_ctl.isolate_process, 
+                                isolate_network: cell.spec().iso_ctl.isolate_network, 
+                            })
+                        .map(|cell| CellWithChildren { cell: Some(cell), children: vec!() })
+                        .collect_vec()
```
future-highway, line 264: This just calls `collect` in a fn that has a `Vec` return type, giving the compiler the required hint for what is being collected. But, since the `cells` field is being set here, the compiler should have enough data and `collect` should be enough. Alternatively, this should be equivalent to `.collect::<Vec<_>>()`

api/v0/cells/cells.proto
```diff
@@ -143,6 +145,17 @@ message CellServiceStopRequest {
...
 message CellServiceStopResponse {}
 
+message CellServiceListRequest {}
+
+message CellServiceListResponse {
+  repeated CellWithChildren cells = 1;
+}
+
+message CellWithChildren {
+  Cell cell = 1;
```
future-highway, line 155: I think we need to consider what data we want back. It seems weird to me to differ on what we want back in Allocate versus List. I don't think you should worry about it in this PR though, we can take care of it in a follow up PR.

@krisnova

auraed/src/cells/cell_service/cell_service.rs
future-highway: That's a name collision. One is `auraed::cells::cell_service::cells::cell` and the other is `Cell` as defined in the proto.

auraed/src/cells/cell_service/cell_service.rs
dmah42: i think i'd pull the map closure out to a function so it can be tested in isolation too.

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: Cool, I was planning to use that same pattern. I was not sure how to test this, though. Maybe I'm not fully understanding the nested `Cell` hierarchy yet. Do we have a way of creating cells within cells?

future-highway: Yes, a cell will be nested or not based on its name. So "my-cell" is a top level cell, and "my-cell/child-cell" is a cell nested in "my-cell".

At the top, `CellService` has a `CellsCache` as implemented by the `Cells` struct. The `Cells` struct has an `Option<CellName>` of its parent `Cell` (`None` for the top `Cells`). When "my-cell" is allocated, it does not have a parent, and neither does the top `CellsCache`, so the allocation (creating the cell) is handled right away. When "my-cell/child-cell" is allocated, the top `CellsCache` sees that the parent names differ, so it proxies to the appropriate immediate child cell (this proxying would continue multiple times if we had, for example, "grandparent/parent/child"; the immediate child of "grandparent" being "parent"). Need to finish this comment later, I'm losing power

auraed/src/cells/cell_service/cell_service.rs
JeroenSoeters: I've extracted it for now. Will probably extract this into separate functions when I have the nested cells in there as well.

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: Thanks, this is super useful! Will add a test case for the nested cells to drive this implementation.

auraed/src/cells/cell_service/cells/cell_graph.rs
```diff
@@ -0,0 +1,20 @@
+use super::CellInfo;
+
+#[derive(Clone)]
+pub struct GraphNode {
+    pub cell_info: Option<CellInfo>,
+    pub children: Vec<GraphNode>,
+}
+
+impl GraphNode {
+    pub fn with_cell_info(&self, cell_info: &CellInfo) -> GraphNode {
```
future-highway, line 10: This is really easy to use incorrectly. Instead of `&self`, the fn should take ownership of `self`.

auraed/src/cells/cell_service/cells/cell.rs
```diff
@@ -229,6 +233,28 @@ impl CellsCache for Cell {
...
+        let cell_info = CellInfo {
+            cell_name: self.cell_name.clone(),
+            cpu: cgroup.cpu,
+            cpuset: cgroup.cpuset,
+            isolate_network: iso_ctl.isolate_network,
+            isolate_process: iso_ctl.isolate_process,
+        };
+
+        let CellState::Allocated { children, .. } = &self.state else {
+            return GraphNode {
```
future-highway, line 250: Instead of returning a node without children, this should result in `Err(CellsError::CellNotAllocated { ... })`. The only way to hit this code is when the cache has gotten out of sync with the system. A call to `Cell` that results in `CellNotAllocated` can be used by `Cells` as a signal to remove the `Cell` from the cache and then handle the situation appropriately. For `cell_graph` I think the appropriate action is to continue to the next cell after removing the current one from the cache.

auraed/src/cells/cell_service/cells/cell_info.rs
```diff
@@ -0,0 +1,12 @@
+use super::cgroups::cpu::CpuController;
+use super::cgroups::cpuset::CpusetController;
+use super::CellName;
+
+#[derive(Clone)]
+pub struct CellInfo {
```
future-highway, line 6: This looks like `CellSpec` + `CellName`. Do we need a dedicated struct for it?

JeroenSoeters: Good catch. Would it make sense to add the `CellName` to the `CellSpec` or should I replace the `CellInfo` struct with a tuple?

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: Just double checking here, I didn't end up using `proxy_if_needed` macro I realized as it looked like the `CellsCache` contained the entire hierarchy.

auraed/src/cells/cell_service/cells/cell_info.rs
future-highway: I think it would be better as a tuple

auraed/src/cells/cell_service/cells/cell.rs
future-highway: It does. Your approach was surprising to me, and just eyeballing it, I was doubting if it works (which is why I tried to test it with more deeply nested cells and led me to find the bug that I had in CellName::to_child). If your approach works, then I don't think you need proxy_if_needed.

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: "out of sync" as in the `cgroup` doesn't exist anymore?

JeroenSoeters: Or are there other failure modes? Kind of want to write a test for this to ensure invalid cells get properly removed in this scenario. Do we have any examples of such tests?

future-highway: It more likely means the cgroup never existed. If you look at `allocate` on `Cells`, you can see that we insert a `Cell` in to the cache and then call `cell.allocate()?`, which could fail. If it fails, then you have a cell in the cache that is not in the `Allocated` state.

future-highway: Also in `Cells`, you can see how `get` and `get_mut` react when encoutering `CellNotAllocated` (they remove the cell from the cache).

JeroenSoeters: Thinking about this a little more. If `cells.allocate()?` fails, wouldn't it make sense to remove that cell from the cache immediately after that call in `Cells::allocate`?

auraed/src/cells/cell_service/cells/cell.rs
```diff
@@ -234,12 +234,9 @@ impl CellsCache for Cell {
...
 
-    fn cell_graph(&self, node: GraphNode) -> GraphNode {
-        let CellState::Allocated { children, .. } = &self.state else {
-            return GraphNode {
-                cell_info: Some((self.cell_name.clone(), self.spec.clone())),
-                children: vec!(),
-            };
+    fn cell_graph(&mut self, node: GraphNode) -> GraphNode {
+        let CellState::Allocated { children, .. } = &mut self.state else {
+            return GraphNode::Invalid { cell_name: self.cell_name.clone() };
```
future-highway, line 239: Looks like you are trying to create an error case. You should have the function's return be a `Result<GraphNode, CellsError>` instead. See the functions above.

auraed/src/cells/cell_service/cells/cell_graph.rs
```diff
@@ -1,24 +1,46 @@
...
+            },
+            GraphNode::Invalid { cell_name: _ } => self,
+        }
+    }
+
+    pub fn unwrap_valid(self) -> Valid {
+        if let GraphNode::Valid { cell_info, children } = self {
+            Valid { cell_info, children }
+        } else {
+            panic!("invalid graphnode")
```
future-highway, line 43: This and the other panics will need to be eliminated. AKA, if we can't handle the error, the error should flow back up to the client.

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: The thing is I actually need the `CellName` to be able to purge it from the cache, that's why I introduced this enum.

JeroenSoeters: https://github.com/JeroenSoeters/aurae/blob/list-cells/auraed/src/cells/cell_service/cells/cells.rs#L250-L265

auraed/src/cells/cell_service/cells/cell.rs
future-highway: If I remember correctly, I fought the borrow checker and lost trying to do that and then didn't go back to try again :).

auraed/src/cells/cell_service/cells/cell.rs
future-highway: CellsError::CellNotAllocated has a `cell_name` field

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: Added a test for a triple nested cell and that seems to work just fine.

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: Ha ok. I left a comment for now to remind us in the future. I spent some time trying to write a test that would cause `allocate()?` to fail, but I was not able to do so. 😒

auraed/src/cells/cell_service/cells/cell.rs
JeroenSoeters: Refactored it again to a `Result` type and it looks a lot cleaner now. Thanks!

auraed/src/cells/cell_service/cells/cell_graph.rs
```diff
@@ -0,0 +1,24 @@
...
+        cell_spec: &CellSpec,
+    ) -> GraphNode {
+        GraphNode {
+            cell_info: Some((cell_name.clone(), cell_spec.clone())),
+            children: self.children,
+        }
+    }
+
+    pub fn with_children(self, children: Vec<GraphNode>) -> GraphNode {
+        GraphNode { cell_info: self.cell_info.as_ref().cloned(), children }
```
future-highway, line 22: The function owns self, so you shouldn't need to clone `cell_info`.

auraed/src/cells/cell_service/cells/cell_name.rs
```diff
@@ -66,6 +66,15 @@ impl CellName {
     pub fn random_for_tests() -> Self {
         Self(PathBuf::from(format!("ae-test-{}", uuid::Uuid::new_v4())))
     }
+
+    #[cfg(test)]
+    pub fn random_nested_for_tests(parent: &CellName) -> Self {
+        Self(PathBuf::from(format!(
+            "{}/ae-test-{}",
+            parent.to_string(),
```
future-highway, line 74: `.to_string()` isn't needed

auraed/src/cells/cell_service/cells/cell_graph.rs
```diff
@@ -0,0 +1,24 @@
...
+pub struct GraphNode {
+    pub cell_info: Option<(CellName, CellSpec)>,
+    pub children: Vec<GraphNode>,
+}
+
+impl GraphNode {
+    pub fn with_cell_info(
+        self,
+        cell_name: &CellName,
+        cell_spec: &CellSpec,
```
future-highway, line 13: The function requires ownership of `cell_name` and `cell_spec`, so shouldn't hide that, IMO. I would clone at the call location if needed.

auraed/src/cells/cell_service/cells/cells.rs
```diff
@@ -243,6 +247,25 @@ impl CellsCache for Cells {
...
+            .values_mut()
+            .map(|c| c.cell_graph(node.clone()))
+            .partition_map(|r| match r {
+                Ok(n) => Either::Left(n),
+                Err(e) => Either::Right(e),
+            });
+
+        for node in invalid {
+            if let CellsError::CellNotAllocated { cell_name } = node {
+                _ = self.cache.remove(&cell_name);
```
future-highway, line 263: I didn't even know you can have the `_ = ...` without let. Can we put the let in? My IDE complains.

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -326,4 +366,93 @@ impl cell_service_server::CellService for CellService {
...
+            .is_ok());
+
+        let result = service.list().await;
+        assert!(result.is_ok());
+
+        let list = result.unwrap();
+        assert_eq!(list.cells.len(), 2);
+
+        let expected_root_cell_names =
+            vec![&parent_cell_name, &nested_cell_name].sort();
```
future-highway, line 419: Should this be `cell_without_children_name` instead of `nested_cell_name`?

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -326,4 +366,93 @@ impl cell_service_server::CellService for CellService {
...
+        let expected_nested_cell_names = vec![&nested_cell_name];
+        let actual_nested_cell_names = parent_cell
+            .unwrap()
+            .children
+            .iter()
+            .map(|c| c.cell.as_ref().unwrap().name.as_str())
+            .collect_vec();
+        assert_eq!(actual_nested_cell_names, expected_nested_cell_names);
+
+        _ = service.free_all().await;
```
future-highway, line 443: Is this call actually needed? It shouldn't be, so needing it indicates a bug somewhere.

examples/cells.ts
```diff
@@ -57,6 +58,33 @@ let started = await cells.start(<runtime.CellServiceStartRequest>{
...
+            weight: 2, // Percentage of CPUs
+            max: 400 * (10 ** 3), // 0.4 seconds in microseconds
+        }),
+    })
+});
+helpers.print(nested_allocated)
+
+// [ Start nested ]
+let nested_started = await cells.start(<runtime.CellServiceStartRequest>{
+    cellName,
```
future-highway, line 75: Do you mean to put the exe in the nested cell?

auraed/src/cells/cell_service/cell_service.rs
JeroenSoeters: It is not. As we were doing the free-ing in the tests for `Cells` I copied that pattern here as I was trying to figure out why `make test-all` was flaky. Will remove this again as this didn't actually solve the problem.

auraed/src/cells/cell_service/cell_service.rs
JeroenSoeters: It should! I hadn't realized that `.sort()` doesn't actually return the sorted vector so I was comparing `()` with `()` here. 🤦‍♂️

examples/cells.ts
JeroenSoeters: I didn't put a whole lot of thought into this example and I'm still not sure if we should use this example to demonstrate the list functionality. I will remove the nested exe for the sake of simplicity.

dmah42 (approved):

PR #326, opened 2023-01-27
Update makefile for more efficient cargo use

future-highway:
Updated the makefile taking into consideration the following (also in the makefile):

- make will not run the same step multiple times
- The ideal order for cargo to reuse artifacts is build -> lint -> test
- Different cargo target variants (nightly is a variant) do not produce compatible artifacts
- Cargo's `install` artifacts are not usable for build, lint, or test (and vice versa)

This PR also updates the workflows affected:
- 001, runs build -> lint -> test. Was lint -> test -> install
- 050, uses build instead of install which should speed it up

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,10 +44,17 @@ export GIT_PAGER = cat
 
 # Aliases
 
+# Keep all as the first command to have it be the default as per convention
 .PHONY: all
 all: install ## alias for install
 
 #------------------------------------------------------------------------------#
+
+# Notes:
+# - make will not run the same step multiple times
+# - The ideal order for cargo to reuse artifacts is build -> lint -> test
+# - Different cargo target variants (nightly is a variant) do not produce compatible artifacts
+# - Cargo's `install` artifacts are not usable for build, lint, or test (and vice versa)
 
 # Super commands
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -164,6 +168,14 @@ auraed-test: musl proto
 auraed-test-all: musl proto
 	@sudo -E $(cargo) test --target $(uname_m)-unknown-linux-musl -p auraed -- --include-ignored
 
+.PHONY: auraed-build
+auraed-build: musl proto
```
jesse-peters, line 172: Would it be beneficial to combine `musl` and `proto` into an `init` step or something?

.github/workflows/001-cargo-install-ubuntu-make-test.yml
```diff
@@ -93,9 +93,9 @@ jobs:
         run: |
           sudo apt-get update
           sudo apt-get install -y musl-tools
-      - name: Cargo Install [make install]
+      - name: Cargo Install [make test]
```
dmah42, line 96: should we just call it what it is? (make test)

Makefile
dmah42: probably not. we can, but it's better to use targets as "build tasks" and we wouldn't (I think) run musl+proto without doing other things.

dmah42 (approved):

jesse-peters (approved):

jesse-peters (approved):

PR #332, opened 2023-01-30
Remove ignore requirement for tests needing root privileges

future-highway:
We've been marking tests that need sudo (root privileges) with ignore, so that we can run the tests in CI.

I noticed the Nix crate had some helpful macros, circumventing the need to use ignore. They aren't part of their public api, so I created a `test-helpers` crate for us and copied the 2 macros we needed to skip when root isn't available.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,7 @@ message      ?=  Default commit message. Aurae Runtime environment.
 cargo         =  cargo
 oci           =  docker
 ociopts       =  DOCKER_BUILDKIT=1
+uid           =  $(shell id -u)
 uname_m       =  $(shell uname -m)
 cri_version   =  release-1.26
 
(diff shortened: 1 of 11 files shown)


future-highway:
Marked the PR as draft until I update it to run the root requiring tests in 035

Makefile
```diff
@@ -357,6 +357,9 @@ alpine: ## Creat an "alpine" OCI bundle in target
 
 # CI Commands
 
+.PHONY: ci-test
+ci-test: pki config test
```
dmah42, line 361: i'm not a huge fan (personal opinion) of having make targets just for convenience.  if the GHA should run `make pki config test` then let's do that so it's explicit what it's doing.  it keeps the local development workflow simple as `make test` should always pass and i don't need to think about what `make ci-test` is or why it's different.

auraed/src/cells/cell_service/cells/cell.rs
```diff
@@ -244,11 +244,12 @@ impl Drop for Cell {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use test_helpers::*;
 
-    // Ignored: requires sudo, which we don't have in CI
-    #[ignore]
     #[test]
     fn test_cant_unfree() {
+        skip_if_not_root!("test_cant_unfree");
```
dmah42, line 251: <3 this

dmah42 (approved):

future-highway:
Closed in favor of #338

PR #336, opened 2023-01-31
Introduce POSIX Signal API

JeroenSoeters:
This PR will introduce a method to the `Observe` API that returns a stream of all [signals](https://man7.org/linux/man-pages/man7/signal.7.html) _handled_ on the host.


auraed/src/observe/mod.rs
```diff
@@ -117,4 +128,64 @@ impl ObserveService for ObserveServiceServer {
...
+    ) -> Result<Response<Self::GetPosixSignalsStreamStream>, Status> {
+        todo!()
+    }
+}
+
+// TODO (jeroensoeters) move this elsewhere once the eBPF code is in
+#[derive(Clone, Debug, PartialEq)]
+struct Signal {
+    //TODO (jeroensoeters) should we introduce an enum here?
+    signal: i32,
```
dmah42, line 147: https://docs.rs/signal/0.5.1/signal/enum.Signal.html ?

Makefile
```diff
@@ -281,6 +281,10 @@ aer-release: $(GEN_RS) $(GEN_TS) aer-lint aer-test ## Lint, test, and install ae
 
 # Commands for other crates
 
+.PHONY: ebpf
+ebpf:
+	$(cargo) xtask build-ebpf
```
dmah42, line 286: i'm not sure the alias is adding much here.. explicit is nice :)

is it useful for local dev to have it? i assume not if we can `make ebpf`.

aurae-ebpf-shared/src/lib.rs
```diff
@@ -0,0 +1,8 @@
+#![no_std]
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct Signal {
+    pub signr: i32,
```
dmah42, line 6: should we use a predefined enum here? https://docs.rs/signal/0.5.1/signal/enum.Signal.html exists and i assume it supports no_std...

Makefile
JeroenSoeters: I started with `xtask` but can probably get rid of it now it's in the `Makefile`. Will have to clean this up anyways as we should probably have the `auraed` build depend on the `ebpf` target.

aurae-ebpf-shared/src/lib.rs
JeroenSoeters: Unfortunately we cannot use that from the kernel. The `Signal` crate depends on `nix` which pulls in a whole slew of transitive dependencies which cause the `aurae-ebpf` build to fail. We could have `load_tracepoint<T>` or the `PerfEventListener<T>` take an extra type parameter, which could be a more idiomatic `Rust` type that would implement `From` for the C-struct that has to traverse the FFI boundary but then again in this codepath we'd be immediately mapping it again to a `proto` type.

aurae-ebpf/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "aurae-ebpf"
```
future-highway, line 2: Would we only every have a single ebpf program, or would we have 1 per task. If it is the latter, then the crate name should be more specific.

auraed/src/logging/log_channel.rs
```diff
@@ -38,8 +38,8 @@ use tokio::sync::broadcast::{self, Receiver, Sender};
 /// Abstraction Layer for one log generating entity
 /// LogChannel provides channels between Log producers and log consumers
 #[derive(Clone, Debug)]
-pub(crate) struct LogChannel {
```
future-highway, line 41: Is this struct being used outside the crate?

auraed/src/lib.rs
```diff
@@ -278,11 +295,20 @@ impl AuraedRuntime {
...
         let discovery_service = DiscoveryService::new();
         let discovery_service_server =
-            DiscoveryServiceServer::new(discovery_service.clone());
+            DiscoveryServiceServer::new(discovery_service);
         health_reporter
             .set_serving::<DiscoveryServiceServer<DiscoveryService>>()
             .await;
 
+        let observe_service = ObserveService::new(
+            Arc::new(LogChannel::new(String::from("TODO"))),
```
future-highway, line 304: Is the `Arc` needed? `LogChannel` can be cloned.

xtask/src/build_ebpf.rs
```diff
@@ -0,0 +1,63 @@
...
+    // Command::new creates a child process which inherits all env variables. This means env
+    // vars set by the cargo xtask command are also inherited. RUSTUP_TOOLCHAIN is removed
+    // so the rust-toolchain.toml file in the -ebpf folder is honored.
+
+    let status = Command::new("cargo")
+        .current_dir(dir)
+        .env_remove("RUSTUP_TOOLCHAIN")
+        .args(&args)
+        .status()
+        .expect("failed to build bpf program");
```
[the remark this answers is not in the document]
JeroenSoeters: I'm getting rid of `xtask` altogether. I just started off with an `aya` template and moved that over to our codebase. I don't see any value keeping `xtask`, and as long as we're targeting `x86_64-*` we'll always have the little endian target (`bpfel-*`).

aurae-ebpf/Cargo.toml
JeroenSoeters: So the way I was thinking about it is having a single binary that would include all the eBPF programs. The programs are named via this [aya attribute](https://github.com/JeroenSoeters/aurae/blob/posix-signals-api/aurae-ebpf/src/main.rs#L17). The `BpfLoader` will then [load these programs](https://github.com/JeroenSoeters/aurae/blob/posix-signals-api/auraed/src/ebpf/loader.rs#L42) from the elf file `aurae-ebpf` into the kernel and [attach them to the correct trace event](https://github.com/JeroenSoeters/aurae/blob/posix-signals-api/auraed/src/ebpf/loader.rs#L48).  I will rename this `name` parameter on `load_tracepoint` to `prog_name` to make it more explicit that it's referring to an actual eBPF program instead of being some random name.

xtask/src/build_ebpf.rs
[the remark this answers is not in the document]
JeroenSoeters: Should we just build it in the root target folder, btw? Not sure what's idiomatic here since it is not a member of the workspace.

auraed/src/lib.rs
JeroenSoeters: I don't think that `LogChannel` is actually used. I just did the minimum to make it compile 😅

xtask/src/build_ebpf.rs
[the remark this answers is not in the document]
future-highway: What is the reasoning for not having it as part of the workspace?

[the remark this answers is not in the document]
JeroenSoeters: I'm not sure why but a lot of extra stuff was ending up in the binary when I included it in the workspace. I think it was like 10-20x times the size and `aya` wasn't able to perform the relocations on the `elf` file (the binary). I tried a few things but then I gave up as excluding it was easier. :) If you want to take a look and see what's going on, you could compile it as a workspace member and inspect the output file with `objdump -S <filename>`. All the names are mangled though so it can be a little hard on the eyes.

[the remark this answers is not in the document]
future-highway: If it can't be part of the workspace that is fine. It should have its own target directory.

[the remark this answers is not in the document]
JeroenSoeters: Cool, `xtask` is gone now and the `Makefile` updated. Note that we do have a dependency on `nightly` now just for this project only as well as a dependency on `bpf-linker`. Once this PR is in I'll update the docs in the `aurae.io` repo.

auraed/src/logging/log_channel.rs
JeroenSoeters: We wouldn't be able to expose it in `ObserverService::new` if it is crate-private that's why I made the change. Before the `ObserveService` wasn't actually used.

dmah42 (changes requested):
putting this here as a merge blocker until after the weekend (live demo for FOSDEM)

dmah42 (commented):
unblocking now the talk is done.

auraed/Cargo.toml
```diff
@@ -65,6 +69,7 @@ procfs = "0.14.2"
 rtnetlink = "0.11.0"
 serde_json.workspace = true
 serde.workspace = true
+signal = "0.5.1"
```
future-highway, line 72: Is this dependency being used? There is a newer version of it also, is there a reason to use the older version? If we only want a signal enum, nix has one at `nix::sys::signal::Signal`

auraed/src/lib.rs
```diff
@@ -268,6 +274,17 @@ impl AuraedRuntime {
 
         // Initialize the bundler
 
+        // Install eBPF probes
+        info!("Installing eBPF probes");
+
+        let bpf_loader = &mut BpfLoader::new()?;
+        let signals = bpf_loader.load_tracepoint::<Signal>(
```
future-highway, line 281: I'd prefer is we had a fn on `BpfLoader` named `load_signals_tracepoint`

auraed/src/observe/mod.rs
```diff
@@ -43,19 +45,29 @@ use tokio_stream::wrappers::ReceiverStream;
...
-use crate::logging::log_channel::LogChannel;
+use crate::{
+    ebpf::perf_event_listener::PerfEventListener,
+    logging::log_channel::LogChannel,
+};
 
 /// The server side implementation of the ObserveService subsystem.
 #[derive(Debug)]
-pub(crate) struct ObserveServiceServer {
+pub(crate) struct ObserveServiceServer {}
```
future-highway, line 55: We can remove this.

auraescript/src/lib.rs
```diff
@@ -93,6 +94,7 @@ fn stdlib() -> Vec<OpDecl> {
     let mut ops = vec![];
     ops.extend(cells::op_decls());
     ops.extend(discovery::op_decls());
+    ops.extend(observe::op_decls());
```
future-highway, line 97: Nit pick: can we make this alphabetical too (cri is also in the wrong spot)

docs/build.md
```diff
@@ -17,12 +17,15 @@ commands provided below.
...
 - [musl libc](https://musl.libc.org)
+- [BPF Linker](https://github.com/aya-rs/bpf-linker)
+
 
 ##### Ubuntu
 
 ```bash
 sudo apt-get install -y protobuf-compiler; # Protocol Buffer Compiler
 sudo apt-get install -y musl-tools; # musl libc
+sudo apt-get install -y build-essentials; # gcc compiler
```
future-highway, line 28: I have concerns about licensing issues and want to make sure we get a review from @krisnova to see if this is ok to add before we merge anything.

future-highway (changes requested):
Blocking until licensing concerns are alleviated.

In my very short amount of research, it seems like eBPF programs declare their license and the kernel will only expose things that are compatible with such license. Diving into aya (which aurae-ebpf) is build on), it seems like it trys to parse a license and if it fails to parse, the [default is gpl](https://github.com/aya-rs/aya/blob/7868fffd7f3fd76611d87d5a0385dbd59ad72f48/aya-obj/src/obj.rs#L488). That gives me concerns. Is there a way to test that the correct license is being parsed?

There is also some guidance on licensing [in the kernel docs](https://docs.kernel.org/bpf/bpf_licensing.html).

docs/build.md
dmah42: not trying to resolve as that's up to @krisnova, but the use of GPL for eBPF is a well-known thing and other projects that embed eBPF programs end up dual-licensed which is _fine_ but a bit of a pain.

can we just pass in a license?

auraed/src/lib.rs
JeroenSoeters: I was hoping to keep the `BpfLoader` agnostic of which actual programs are loaded and just supply it with a couple of methods like `load_tracepoint`, `load_kprobe`, `load_uprobe` etc and maintain a list of the actual programs we want to load elsewhere. I can add an extra function for now until we figure out where this list would live.

auraed/src/logging/log_channel.rs
future-highway: We'll refactor in a future PR

aurae-ebpf-shared/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "aurae-ebpf-shared"
+version = "0.1.0"
+edition = "2021"
```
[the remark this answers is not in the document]
future-highway: Typo "publish", not "package"

xtask/src/build_ebpf.rs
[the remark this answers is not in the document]
future-highway: This repo is also used for aurae.io. The other repo is archived.

auraed/src/lib.rs
future-highway: I understand your reasoning and would agree if this were library code, but since this is internal I think the helper methods are the right approach. Thanks for making the change.

docs/build.md
JeroenSoeters: The licensing pertains to BPF helper functions in the kernel, [some of them are GPL-ed and some are not](https://github.com/iovisor/bcc/blob/master/docs/kernel-versions.md#helpers). This PR is using [`BPF_FUNC_perf_event_output()`](https://github.com/JeroenSoeters/aurae/blob/posix-signals-api/aurae-ebpf/src/main.rs#L44) which has a GPL license. I think we would need to dual license only the BPF code [like Cilium does](https://github.com/cilium/cilium#license).

xtask/src/build_ebpf.rs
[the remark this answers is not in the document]
JeroenSoeters: Yep found it already and included the changes in this PR.

docs/build.md
future-highway: If the function we are using is GPL-ed and everything was working (which I assume is true given the discord post), then **`aya` must be assuming we are licensed under GPL**.

Some food-for-thought:
- If some code is going to be GPL-ed, should we put it in a separate repo with its own builds and releases? Even if it doesn't make a technical difference, I think that is a better approach then what I assume is the alternative -- update the license files to say "this is the default unless otherwise indicated".
- Do we need to update the CLA?

Also, I want to clarify that I'm not trying to suggest the direction we go in; I don't have the knowledge needed to make those suggestions.

auraescript/src/lib.rs
```diff
@@ -67,8 +67,8 @@ mod builtin;
 mod cells;
 mod cri;
 mod discovery;
-mod health;
 mod observe;
+mod health;
```
future-highway, line 71: Nitpick: alphabetical

Sidenote: I think rustfmt would take care of this automatically, so you may not have it turned on.

Also, don't think of the nitpicks as blockers. I wouldn't block the merge just for those.

auraed/src/lib.rs
```diff
@@ -268,6 +273,12 @@ impl AuraedRuntime {
 
         // Initialize the bundler
 
+        // Install eBPF probes
+        info!("Installing eBPF probes");
+
+        let bpf_loader = &mut BpfLoader::new()?;
+        let signals = bpf_loader.load_signals_tracepoint()?;
```
future-highway, line 280: This seems like it loads the bpf program in all contexts (PID 1, nested, container, etc.). Is that correct?

@krisnova or @dmah42

docs/build.md
JeroenSoeters: > aya must be assuming we are licensed under GPL.

It's not just `aya`, you literally have to tell the kernel that the BPF program you're loading is GPL licensed. This is what that `syscall` looks like:

```bpf(BPF_PROG_LOAD, {prog_type=BPF_PROG_TYPE_TRACEPOINT, insn_cnt=31, insns=0x5555593ae5c0, license="GPL", log_level=0, log_size=0, log_buf=NULL, kern_version=KERNEL_VERSION(6, 1, 8), prog_flags=0, prog_name="signals", prog_ifindex=0, expected_attach_type=BPF_CGROUP_INET_INGRESS, prog_btf_fd=0, func_info_rec_size=0, func_info=NULL, func_info_cnt=0, line_info_rec_size=0, line_info=NULL, line_info_cnt=0, attach_btf_id=0, attach_prog_fd=0, fd_array=NULL}, 144)```

I'm not sure if we would want to create a separate repo for the BPF code, though. We'd have to run a private crate registry, and we'd be introducing a lot of friction in the development process, not to mention that for every change, we would now need 2 PRs instead of 1.

If you look at some of the other popular OSS eBPF projects like Cilium and Falco, they typically have a mention somewhere of which files are dual licensed and have a different license header for those files.

https://github.com/cilium/cilium#license
https://github.com/falcosecurity/libs/blob/master/NOTICES

File: auraed/src/lib.rs (hunk 15 above)
Comment (JeroenSoeters) [reply auraed/src/lib.rs:280]: Yea I was thinking about this as well, I am going to change this to only load the BPF programs in the host daemon for now. I'm actually not fully understanding the `observe` gRPC endpoint tbh. Is there a way to `observe` a nested `auraed` right now? If there is we'd anyways need to think about mapping the host `PID` to the namespaced `PID` in cases where the `PID namespace` is `unshared`.

Comment (JeroenSoeters) [reply auraed/src/lib.rs]: Just changed this. eBPF is only loaded from the host daemon for now.

Comment (JeroenSoeters) [reply auraed/src/lib.rs]: Left some comments [on this issue](https://github.com/aurae-runtime/aurae/issues/366) on how we could `observe` signals for nested auraed processes. Probably should do this in a different PR though.

File: Makefile [hunk 16]
```diff
@@ -221,6 +230,16 @@ auraed-start: ## Starts the installed auraed executable
...
 
+.PHONY: ebpf-build ## Build auraed eBPF probes (debug)
+ebpf-build: nightly bpf-linker
+	cd ./aurae-ebpf && \
+		$(cargo) +nightly build --target=bpfel-unknown-none -Z build-std=core
+
+.PHONY: ebpf-release ## Build auraed eBPF probes
+ebpf-release: nightly bpf-linker
+	cd ./aurae-ebpf && \
+		$(cargo) +nightly build --package aurae-ebpf --target=bpfel-unknown-none -Z build-std=core --release
```
dmah42, line 241: is this different to `$(cargo) +nightly build --path ./aurae-ebpf --package ...` ?

docs/build.md
dmah42: dual-licensing may not be an issue for our use-case as we're not shipping this as a library to be integrated into other projects.  ie, anyone _using_ aurae is not going to be subject to the terms in GPL unless they're building code linking aurae in, which is not what i'd expect.

but IANAL and i agree @krisnova should have final say.  also, i don't feel qualified to assess the eBPF parts of this PR :)

Makefile
JeroenSoeters: I'm not showing `--path` as a valid `cargo build` flag. I'm on cargo 1.66.1.

```error: unexpected argument '--path' found```

auraescript/src/lib.rs
JeroenSoeters: I do have `rustfmt` enabled on save so I'm a bit puzzled as well as why it doesn't order these automatically.

Makefile
dmah42: ah looks like we use install for other executables.

krisnova:
Hey everyone jumping in I am going to look at the code now and possibly make some superficial changes to get this merged in!

krisnova:
Also we will need all contributors to this PR to add a GitHub comment with the following language:

> I hereby agree to the terms and conditions in the GNU Public License (GPL) Version 2 for all work contributed.

JeroenSoeters:
I hereby agree to the terms and conditions in the GNU Public License (GPL) Version 2 for all work contributed.

krisnova (approved):

PR #337, opened 2023-01-31
Rename and provide guidance on workflow naming

future-highway:
Made the workflow numbering a little less random.


jesse-peters (changes requested):
Looks good. You'll need to change the images too in the README, and for the docs as well I think.

jesse-peters (approved):

PR #338, opened 2023-01-31
Remove ignore requirement for tests needing root privileges

future-highway:
Opening in favor of #332 as fork based PRs can't change images


jesse-peters (approved):
I can't give you a great review of the rust changes, but the rest looks good to me.

PR #339, opened 2023-01-31
Don't lint docs again when deploying

future-highway:


jesse-peters (approved):
Love the new cache numbering

PR #340, opened 2023-02-01
Docker config for git safedirs

jesse-peters:


future-highway (approved):
The fix looks right. Admittedly, I don't know the security implications. It seems like we are trusting the doc generators, which I think is an acceptable risk. We are already trusting every rust crate anyway.

jesse-peters:
It seems as though since the container isn't really leaving our control we should be ok. We're generating that code anyways

PR #345, opened 2023-02-01
Generalize Cells iteration

future-highway:
Made `CellService` use a more general `get_all` implemented on `CellsCache`.


auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -39,6 +38,7 @@ use super::{
     },
     Result,
 };
+use crate::cells::cell_service::cells::CellsError;
```
dmah42, line 41: what's the difference between super::cells and crate::cells?

dmah42 (approved):

auraed/src/cells/cell_service/cell_service.rs
future-highway: `crate` is from the crate's root, and `super` is relative to the current module (i.e., `crate::cells::cell_service`). Now that we don't have `runtime`, we can probably flatten the structure a bit as I was also having trouble keeping the track of the module paths.

PR #346, opened 2023-02-01
Don't run `buf generate` if api directory is unchanged

future-highway:
Currently, our makefile will run `buf generate` even when the files in the api directory are unchanged. This PR prevents `buf generate` from running when nothing has changed by leveraging a rust build.rs and cargo's rerun hints.

The benefit is that for local development, crates that depend on the output of `buf generate` will have a better chance of caching their builds.

The drawback is that if someone alters the generated files, `buf generate` won't run to revert the changes. They would need to run `make clean` first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -410,6 +410,10 @@ dependencies = [
 ]
 
 [[package]]
+name = "buf-buffer"
+version = "0.0.0"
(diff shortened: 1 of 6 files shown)


future-highway:
#348

PR #348, opened 2023-02-02
ensure we only run buf when we need to

dmah42:
this change creates some pattern matches between the input protos and the expected outputs (rust and typescript).  it then defines a rule that runs buf if any of the outputs are missing.

this means that make will only run buf when necessary and, because the first run creates all the files, will only run it once.


dmah42:
docs failure is due to the "sub-targets" of docs not having dependencies correctly set.

dmah42:
udeps: a "make proto" escaped me

future-highway (approved):
closes #346

PR #352, opened 2023-02-04
ensure aurae is using consistent clap version

dmah42:


future-highway (approved):
If I knew they were that compatible, I wouldn't have left the todo, lol. Thanks for taking care of it.

PR #353, opened 2023-02-04
clean up some lint warnings, removing dead code

dmah42:


auraed/src/init/system_runtimes/pid1_system_runtime.rs
```diff
@@ -113,7 +113,7 @@ impl SystemRuntime for Pid1SystemRuntime {
         trace!("init of auraed as pid1 done");
 
         let socket_addr = socket_address
-            .unwrap_or_else(|| "0.0.0.0:8080".into())
+            .unwrap_or_else(|| DEFAULT_NETWORK_SOCKET_ADDR.into())
```
future-highway, line 116: We have an open issue about using ipv6 as the default. Do we want this to be an ipv6 address?

future-highway (dismissed):

auraed/src/init/system_runtimes/pid1_system_runtime.rs
dmah42: oh sure.. the networking layer we have can create IPv4 or IPv6 routes.  if the ideal is to go for ipv6 only, that's great.

future-highway (approved):

PR #354, opened 2023-02-04
Port to libcgroups crate

future-highway:
This PR uses the libcgroups crate (part of youki repo) instead of cgroups-rs. We avoid issues with musl incompatibilities by not using the default features and only using the v2 feature which seems to compile without issue (I didn't look into which feature causes the issue).

### DIFF
--- a/auraed/Cargo.toml
+++ b/auraed/Cargo.toml
@@ -59,10 +59,8 @@ log = "0.4.17"
 multi_log = "0.1.2"
 netlink-packet-route = "0.13.0" # Used for netlink_packet_route::rtnl::address::nlas definition
 nix = { workspace = true, features = ["sched"] }
-nix_025 = { package = "nix", version = "0.25" } # needed for libcgroups as of v0.0.4
 ocipkg = "0.2.8"
-oci-spec = "0.6.0"
-oci-spec-05 = { package = "oci-spec", version = "0.5" } # needed for libcgroups as of v0.0.4
+oci-spec = "0.5" # upgrade to 0.6 blocked by libcgroups as of v0.0.4
 procfs = "0.14.2"
 rtnetlink = "0.11.0"
 serde_json.workspace = true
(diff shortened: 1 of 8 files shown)


auraed/Cargo.toml
```diff
@@ -53,14 +52,17 @@ ipnetwork = "0.20.0"
...
 libcontainer = { git = "https://github.com/krisnova/youki", branch = "musl-libcontainer", features = ["v2"], default-features = false }
 liboci-cli = "0.0.4"
 log = "0.4.17"
 multi_log = "0.1.2"
 netlink-packet-route = "0.13.0" # Used for netlink_packet_route::rtnl::address::nlas definition
 nix = { workspace = true, features = ["sched"] }
+nix_025 = { package = "nix", version = "0.25" } # needed for libcgroups as of v0.0.4
 ocipkg = "0.2.8"
 oci-spec = "0.6.0"
+oci-spec-05 = { package = "oci-spec", version = "0.5" } # needed for libcgroups as of v0.0.4
```
dmah42, line 65: can we downgrade our workspace version to avoid duplicates?

auraed/src/cells/cell_service/cells/cgroups/cgroup.rs
```diff
@@ -56,89 +59,144 @@ impl Cgroup {
...
+        )
+        .expect("valid cgroup");
 
-        Self { non_leaf, leaf }
+        Ok(Self { cell_name })
     }
 
-    pub fn add_task(&self, pid: Pid) -> cgroups_rs::error::Result<()> {
-        self.leaf.add_task_by_tgid((pid.as_raw() as u64).into())
+    pub fn add_task(&self, pid: Pid) -> anyhow::Result<()> {
```
dmah42, line 141: did you not want to get rid of anyhow?

dmah42 (approved):

PR #357, opened 2023-02-04
Init cargo deny

future-highway:
I'd like to get cargo deny passing and to have a check in CI to make sure it remains so.

The deny.toml file was generated via `cargo deny init` plus a few edits:
- Added MIT and Apache 2.0 to allowed licenses
- Set `confidence-threshold` to 1.0 (we would need manually ok anything else)
- Set `copyleft` to deny

We currently do not pass the licenses check with only MIT and Apache. We also don't pass the advisories check.


deny.toml
```diff
@@ -0,0 +1,269 @@
...
+# https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
+[licenses]
+# The lint level for crates which do not have a detectable license
+unlicensed = "deny"
+# List of explicitly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.11 short identifier (+ optional exception)].
+allow = [
+    "MIT",
+    "Apache-2.0",
```
dmah42, line 106: we could preemptively add BSD3 and other "notice" licenses

dmah42:
do you want to merge this as is, or do you want to first get us passing?

krisnova (approved):

PR #363, opened 2023-02-05
Introduce cgroups v2 memory controller 

dmah42:


api/v0/cells/cells.proto
```diff
@@ -87,12 +88,12 @@ service CellService {
   rpc List(CellServiceListRequest) returns (CellServiceListResponse) {}
 }
 
-/// An Aurae cell is a name given to Linux control groups (cgroups) that also include
-/// a name, and special pre-exec functionality that is executed from within the same context
-/// as any executables scheduled.
+/// An Aurae cell is a name given to Linux control groups (cgroups) that also
```
future-highway, line 91: Nothing to do with this PR, but we need to make these double slashes and think about documenting a formatter. Is there an equivalent to rustfmt.toml for proto files?

api/v0/cells/cells.proto
```diff
@@ -167,6 +164,7 @@ message Cell {
 
   CpuController cpu = 2;
   CpusetController cpuset = 3;
+  MemController mem = 4;
```
future-highway, line 167: Can we make this `MemoryController` and `memory` to match the controller name as seen in the docs?

api/v0/cells/cells.proto
future-highway: To clarify, I don't expect this to be done in this PR. I was just noting my thoughts.

dmah42: i don't know about a rustfmt.toml, but things like prettier will format protos.

api/v0/cells/cells.proto
dmah42: done

api/v0/cells/cells.proto
```diff
@@ -225,3 +221,39 @@ message CpusetController {
...
+  // NOTE: unused by aurae
+  optional int64 min = 1;
+
+  // Best-effort memory protection. If the memory usage of a cgroup is within
+  // its effective low boundary, the cgroup’s memory won’t be reclaimed unless
+  // there is no reclaimable memory available in unprotected cgroups. Above
+  // the effective low boundary (or effective min boundary if it is higher),
+  // pages are reclaimed proportionally to the overage, reducing reclaim
+  // pressure for smaller overages.
+  // NOTE: unused by aurae
```
future-highway, line 243: Low is used. High is not

future-highway (commented):
I can't give a proper review but figured I'd leave a few comments. Generally looks good though.

auraed/src/cells/cell_service/cells/cgroups/cgroup.rs
```diff
@@ -128,8 +130,29 @@ impl Cgroup {
                     cpu_builder
                 };
 
-            let cpu = cpu_builder.build().expect("valid builder");
-            builder.cpu(cpu)
+            let memory_builder = LinuxMemoryBuilder::default();
+            let memory_builder =
+                if let Some(MemoryController { min: _, low, high: _, max }) =
```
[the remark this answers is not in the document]
dmah42: yeah i am surprised by this too. both cgroups-rs and libcgroups only implement partial memory control.  i figured it was useful for us to plumb this through so we can iterate towards more support without changing the API later.

auraed/src/cells/cell_service/cells/cgroups/cgroup.rs
```diff
@@ -88,7 +90,7 @@ impl Cgroup {
         let builder = LinuxResourcesBuilder::default();
 
         // oci_spec, which libcgroups uses, combines the cpu and cpuset controllers
-        let builder = if cpu.is_some() || cpuset.is_some() {
+        let builder = if cpu.is_some() || cpuset.is_some() || memory.is_some() {
```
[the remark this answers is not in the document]
dmah42: i had it separate and got into trouble trying to set cpu and memory on builder as copies started to creep in.  this felt clean too.

api/v0/cells/cells.proto
dmah42: some documentation may be out of date. :)

auraed/src/cells/cell_service/cells/cgroups/memory/mod.rs
```diff
@@ -0,0 +1,39 @@
...
+ *   See the License for the specific language governing permissions and      *
+ *   limitations under the License.                                           *
+ *                                                                            *
+\* -------------------------------------------------------------------------- */
+
+use super::{Limit, Protection};
+
+#[derive(Debug, Clone)]
+pub struct MemoryController {
+    pub min: Option<Protection>,
```
future-highway, line 35: Sorry, I wasn't clear. I think min is an Allocation as it is supposed to be guaranteed. Low is a protection and therefore not guaranteed.

dmah42: in the docs, memory min is described as "Hard memory protection" and low as "Best-effort memory protection".  high is "Memory usage throttle limit" and max is "Memory usage hard limit".

so i think this is correct as written.

future-highway: I agree with your reading, but still think the code is incorrect. Low and min shouldn't be the same type.

future-highway (approved):
Reviewing from my phone but everything looks good. I'll leave it to you dma if you want someone else to also review before merging.

future-highway (dismissed):
Reviewing from my phone but everything looks good. I'll leave it to you dma if you want someone else to also review before merging.

krisnova (approved):

krisnova:
Can we fix the conflicts and then merge this in please?

krisnova (approved):

PR #364, opened 2023-02-05
Port to libcgroups (2nd attempt)

future-highway:
Reverted the revert and fixed cgroup order or operations. Our tests now also set some settings on CpuController so that we test applying constraints. As well, I ran a few of the example auraescript files to make sure they work (I should have done that the first time instead of just using aer).

### DIFF
--- a/auraed/src/cells/cell_service/cells/cgroups/cgroup.rs
+++ b/auraed/src/cells/cell_service/cells/cgroups/cgroup.rs
@@ -28,13 +28,19 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
+use super::Result;
+use crate::cells::cell_service::cells::cgroups::CgroupsError;
 use crate::cells::cell_service::cells::{
     cgroups::{CpuController, CpusetController},
     CellName, CgroupSpec,
 };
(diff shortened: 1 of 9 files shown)


dmah42:
do you solemnly swear the scripts run? :)

dmah42 (approved):

future-highway:
I solemnly sweat cells.ts, cells_output.ts, and cells_quota.ts all ran without error on my computer :).

auraed/src/cells/cell_service/cells/cgroups/cgroup.rs
```diff
@@ -28,117 +28,201 @@
...
-    leaf: cgroups_rs::Cgroup,
+    cell_name: CellName,
 }
 
 impl Cgroup {
-    pub fn new(cell_name: CellName, spec: CgroupSpec) -> Self {
+    pub fn new(
+        cell_name: CellName,
+        spec: CgroupSpec,
+        nested_auraed_pid: Pid,
```
future-highway, line 58: A safer signature would take `nested_auraed: &NestedAuraed` and then have the `let pid = nested_auraed.pid()` in the function, preventing passing a `Pid` from a different process.

PR #367, opened 2023-02-05
Updating some docs

justbaileym:
I was originally just trying to build `auraescript`, which set me on this tangent of updating docs, which now lead me to trying to get the docs working on my Macbook.

These commits and this PR description will hopefully be better once it's all figured out. Comments welcome though!

We have 99% of what's necessary to make sure the committed docs are up-to-date with the generated bits, because we have the lint and build jobs.

The [lint](https://github.com/aurae-runtime/aurae/blob/main/.github/workflows/290-check-website-documentation-aurae-builder-make-docs.yml#L75-L77) checks for typos, so can also check if it's generating any changes or not, by doing a `git diff` and checking the exit code or something.

The [build](https://github.com/aurae-runtime/aurae/blob/main/.github/workflows/291-deploy-website-documentation-aurae-builder-make-docs.yml#L70-L73) job generates the docs, then force-pushes to gh-pages, so the actual docs published will be up-to-date, but not the generated docs stored in the repo.


auraed/README.md
```diff
@@ -40,15 +40,16 @@ OPTIONS:
 
 We suggest using the [aurae](https://github.com/aurae-runtime/aurae) repository for building all parts of the project.
 
-If you intend on building this repository directly you can leverage the Makefile in this repository.
+It is possible to build `auraed` by itself. Check out this repository and use the Makefile.
 
 ```bash
-make
+make auraed-debug
```
dmah42, line 46: `make auraed` should be the right thing to build the debug version.

auraescript/README.md
```diff
@@ -36,11 +36,10 @@ cd aurae
 make pki config all
 ```
 
-Alternatively it is possible to build `aurascript` by itself. Check out this repository and use the Makefile.
+Alternatively it is possible to build `auraescript` by itself. Check out this repository and use the Makefile.
 
 ```bash
-cd auraescript
-make
+make auraescript-debug
```
dmah42, line 42: same .. just `make auraescript`

dmah42:
the workflow i would love to see is where no-one has to think about generating the docs.  this means one of two things:

1. `make` generates the docs automatically, just as it generates the .rs and .ts files today
2. a github action generates and commits the docs on merge to main

i don't have an opinion either way on which option we go with, but we should pick one and make that the One True Way.

justbaileym:
That sounds good.

I think the first option sounds simpler, though at the same time I'm not 100% sure needs to be done for that.

justbaileym:
@cla-bot check

dmah42:
> That sounds good.
> 
> I think the first option sounds simpler, though at the same time I'm not 100% sure what needs to be done for that.

i think it is simpler.  it would be a matter of having some mapping from input proto/rust file to output doc and then a target in make.  then that target should be a prerequisite for testing or installing or something.

an even simpler version would be to have a target in the Makefile that creates the docs from the proto/rust that is a prerequisite for test targets or something.

justbaileym:
Another discovery! The `auraed` crate's docs are missing from gh-pages, because of specifying [`--target`](https://github.com/aurae-runtime/aurae/blob/main/Makefile#L257), so they end up in `target/${MUSL_TARGET}/doc` instead of `target/doc`. 🙃

justbaileym:
I might have veered a little off the track, but I'll look at this some more during the week and figure it out.

Makefile
```diff
@@ -258,6 +258,7 @@ docs-crates: musl $(GEN_TS) $(GEN_RS) ## Build the crate (documentation)
 	$(cargo) doc --no-deps --package auraescript
 	$(cargo) doc --no-deps --package aurae-client
 	$(cargo) doc --no-deps --package aer
+	cp -rv target/$(uname_m)-unknown-linux-musl/doc/* docs/crate
```
dmah42, line 261: well spotted!

docs/auraed/index.md
```diff
@@ -21,7 +21,7 @@ We suggest using the [aurae](https://github.com/aurae-runtime/aurae) repository
 If you intend on building this repository directly you can leverage the Makefile in this repository.
 
 ```bash
-make
+make auraed-debug
```
dmah42, line 24: just `auraed` :)

docs/stdlib/v0/index.md
```diff
@@ -1,458 +0,0 @@
-# Protocol Documentation
```
dmah42, line 1: this is being deleted because it's correctly generated on the fly?

docs/auraed/index.md
justbaileym: My bad. 🙈.

I updated it in the files I deleted, but had already copy-pasted it. 😐 

Oh wait, no it's something different from that, because I was also trying to move it somewhere else, so I don't think that's the reason.

docs/stdlib/v0/index.md
justbaileym: Yeah!

justbaileym:
I think I should maybe break this PR into multiple, because the diff doesn't make it super obvious now what is:

- moving docs around - to be able to more easily generate them and avoid duplication and keep them in sync
- drift between committed and generated docs - which shouldn't be a problem if we don't commit any generated docs
- making sure our generated docs are being created and look correct - to avoid any regressions

dmah42:
> I think I should maybe break this PR into multiple, because the diff doesn't make it super obvious now what is:
> 
> * moving docs around - to be able to more easily generate them and avoid duplication and keep them in sync
> * drift between committed and generated docs - which shouldn't be a problem if we don't commit any generated docs
> * making sure our generated docs are being created and look correct - to avoid any regressions

i'm happy with it in one PR if you are.  i can follow along i think.

justbaileym:
> i'm happy with it in one PR if you are. i can follow along i think.

I think the diff was confusing at one point for me, because I hadn't committed everything, and there was a bit of drift on my local machine, but I'm less paranoid now, after re-generating it all and navigating through the generated docs!

dmah42 (commented):
let me know when you feel it's ready for a proper review :)

krisnova:
ping -- is this ready for review?

justbaileym:
Sorry, needed to fix some stuff on my machine before I could test it one last time. I'll fix the conflicts, then I think it should be all ready for review!

Makefile
```diff
@@ -275,6 +276,11 @@ docs-other:
 docs-serve: docs ## Run the aurae.io static website locally
 	sudo -E ./hack/serve.sh
 
+.PHONY: docs-test
+docs-test: docs/crate/auraed/ docs/crate/aer/ docs/crate/aurae_client docs/crate/auraescript
```
justbaileym, line 280: This should consistently have a slash or no slash and probably test more file paths and directories.

Not sure if anyone has better ideas for how to test they've been generated.

The error message could definitely also be better, telling folks how to generate the docs it's sad are missing instead of just that some are missing.

auraed/src/lib.rs
```diff
@@ -37,6 +38,58 @@
...
+//!
+//! To run auraed as a standard library server you can run the daemon alongside your current init system.
+//!
+//! ```bash
+//! sudo -E auraed
+//! ```
+//!
+//! Additional flags are listed below.
+//!
+//! ```bash
```
krisnova, line 59: We should remove the bash example help as these are documented in other places automatically. We aren't gaining anything by adding it here except creating yet another place we have to manually keep in sync. 

Please remove ALL of the bash help output.

auraed/src/lib.rs
```diff
@@ -37,6 +38,58 @@
...
+//!
+//! We suggest using the [aurae](https://github.com/aurae-runtime/aurae) repository for building all parts of the project.
+//!
+//! It is possible to build `auraed` by itself. Check out this repository and use the Makefile.
+//!
+//! ```bash
+//! make auraed
+//! ```
+//!
+//! or using Cargo directly
```
krisnova, line 85: For now using Cargo manually is unsupported. Please remove it from the docs such that `make` is the only supported way of building Aurae.

auraescript/src/lib.rs
```diff
@@ -27,6 +27,63 @@
...
 \* -------------------------------------------------------------------------- */
+//! # AuraeScript
+//!
+//! AuraeScript is a turing complete language for platform teams built on [Deno](https://deno.land).
+//!
+//! AuraeScript is a lightweight client that wraps the [Aurae Standard Library](https://aurae.io/stdlib/).
+//!
+//! AuraeScript is a quick way to access the core Aurae APIs and follows normal UNIX parlance. AuraeScript should feel simple and intuitive for any Go, C, Python, or Rust programmer.
+//!
+//! ```typescript
```
krisnova, line 38: Please remove all of the typescript examples from this file. We should reference the files in `/examples` for examples of AuraeScript

auraescript/src/lib.rs
```diff
@@ -27,6 +27,63 @@
...
+//!     })
+//! }).then(r => {
+//!     // @ts-ignore
+//!     Deno.core.print("done")
+//! });
+//! ```
+//!
+//! ## Build From Source
+//!
+//! ⚠️ Early Active Development ⚠️
```
krisnova, line 60: Remove this

auraescript/src/lib.rs
```diff
@@ -27,6 +27,63 @@
...
+//! make pki config all
+//! ```
+//!
+//! Alternatively it is possible to build `auraescript` by itself. Check out this repository and use the Makefile.
+//!
+//! ```bash
+//! make auraescript
+//! ```
+//!
+//! ...or manually using Cargo.
```
krisnova, line 74: Remove all references to building with Cargo, we are only supporting `make` at this time

justbaileym:
Sorry, life has been a bit hectic. I'll update this PR this week!

krisnova:
Closing in favor of #459

Issue #371, opened 2023-02-07
Remove `aurae` prefixes


jesse-peters:
We currently use the `aurae` prefix on our folders. We should remove those since that's implied by the fact that they're in the ` aurae` repository.

Below it should read:
`client`
`proto`

<img width="210" alt="image" src="https://user-images.githubusercontent.com/3060031/217294126-bb641f15-ee50-4826-adb9-0c72862afd89.png">

dmah42:
working on proto now. I'll leave that as an example if someone wants to do client.

dmah42:
does aurae-client make sense as a published crate?  i thought with auraed, auraescript, and aer we were covered.

krisnova:
I believe this is now closed. Very important to see my note in https://github.com/aurae-runtime/aurae/pull/372#issuecomment-1426829805

PR #372, opened 2023-02-07
s/aurae-proto/proto/

dmah42:
part of #371 

simplify the view of things by removing the previx.


future-highway (dismissed):
Currently weakly holding the opinion that we shouldn't make this change. But, looking from my phone, I don't see any issues. Once again, I'll leave it you dma if you want a different review as well.

dmah42:
gonna escalate this to @krisnova for BDFL approval.

krisnova:
Curious why @future-highway thinks we shouldn't rename the directory? 

I have a weakly held opinion that we should try not to use the name of the project in nested folder names, but if there is a good reason for it such as "This is the main Aurae API" I am actually okay with it.

dmah42:
proto is not the main API (that's in api) but if we ever publish aurae-proto or aurae-client as crates, we'd want the prefix in there.

i assume we can publish a crate called "aurae-foo" from a directory called "foo" but i don't know for sure.

future-highway:
By convention, the crate name would match the directory name but  I don't believe there is a technical requirement at any point, including publishing a crate. I think it's ok to rename if that is the preference.

dmah42:
ready to merge :)

future-highway (approved):
LGTM

krisnova:
Looked at this on the stream and I think we are okay with have packages names `foo` that are published to crates.io as `aurae-foo`.

Additionally we now have control [crates.io/crates/aurae](https://crates.io/crates/aurae) which maybe is where we want to publish crates such as the client? 

I am wondering under what conditions a developer would try:

```bash 
cargo add aurae
```

and I think maybe the expectation is to get the rust client? although we can make that decision later! 

Hope this helps explain my reasoning. Basically I am convinced we will likely be renaming/mapping these directories to other crate names (and maybe even repos?) in the future.

PR #375, opened 2023-02-07
auraescript: make helpers.print variadic

jcorbin:
- update examples/cells.ts to demonstrate it
- add editor config so that tsserver stops fighting the project's 4-space indent style

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,34 @@
+; https://editorconfig.org/
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+trim_trailing_whitespace = true
+
+# Unix-style newlines with a newline ending every file
+end_of_line = lf
+insert_final_newline = true
+
+[Makefile]
+indent_style = tab
+indent_size = 4
+
+[*.md]
+indent_size = 4
+trim_trailing_whitespace = false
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,20 @@
+# top-most EditorConfig file
```
dmah42, line 1: can we take some from https://github.com/aurae-runtime/ae/blob/main/.editorconfig ?

jcorbin:
> In order to contribute to a Nivenly Foundation project you must [sign and agree to the CLA](https://cla.nivenly.org). Reply with `@cla-bot check` to check again.

@cla-bot check

.editorconfig
jcorbin: done

jcorbin:
I suppose it'd be nicer to use [Deno.inspect()](https://deno.land/api@v1.29.1?s=Deno.inspect) (which formats like `console.log` would do instead of `JSON.stringify` ; or even to question "hey, why aren't we just using console.log anyhow?"
... but a brief attempt to use either ended up with them being undefined
... will have to look into what's going on with the auraescript deno globals soon, just starting to get my footing here

dmah42 (approved):

dmah42:
thanks!

PR #380, opened 2023-02-09
More thoughtful buffer sizes for the eBPF perf event programs.

JeroenSoeters:
As discussed in the stream @krisnova I put some more thought into those buffer sizes. Since the per-CPU buffers of the  BPF_PERF_EVENT_ARRAY BPF map are mmaped it made sense to align those buffers with pages.

Also added some comments in an attempt to make this code a bit more readable.

### DIFF
--- a/auraed/src/ebpf/loader.rs
+++ b/auraed/src/ebpf/loader.rs
@@ -38,6 +38,7 @@ use aya::{
 };
 use bytes::BytesMut;
 use log::info;
+use procfs::page_size;
 use std::mem::size_of;
 use tokio::sync::broadcast;
 use tracing::{trace, warn};
@@ -55,9 +56,8 @@ pub struct BpfLoader {
 const INSTRUMENT_TRACEPOINT_SIGNAL_SIGNAL_GENERATE: &str =
     "instrument-tracepoint-signal-signal-generate";
 
-/// Definition of the channel capacity for the perf event buffer fro ALL CPUs.
-// TODO: We should consider some basic math here. Perhaps 1024 * Available CPUs (Relevant for nested auraed in cgroups)
-const CHANNEL_CAPACITY: usize = 1024;
+/// Size (in pages) for the circular per-CPU buffers that BPF perfbuf creates.
+const PER_CPU_BUFFER_SIZE_IN_PAGES: usize = 2;
 
 impl BpfLoader {
     pub fn new() -> Self {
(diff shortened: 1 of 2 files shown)


auraed/src/ebpf/loader.rs
```diff
@@ -136,7 +166,7 @@ impl BpfLoader {
 
                     if events.lost > 0 {
                         warn!(
-                            "queues are getting full, lost {} perf events",
+                            "queues are getting full, dropped {} perf events",
```
krisnova, line 169: Can we please open up an issue for dropped eBPF events here? I believe we will need to start understanding the conditions in which something like this will happen and see if we can't find ways of getting closer to a guarantee that we do not drop events.

auraed/src/observe/mod.rs
```diff
@@ -174,52 +174,53 @@ impl observe_service_server::ObserveService for ObserveService {
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::ebpf::loader::BpfLoader;
+    use std::{process::Command, time::Duration};
+    use test_helpers::*;
+    use tokio::sync::Mutex;
+
+    #[tokio::test]
+    #[ignore] // Cannot run this in CI as we don't want to install eBPF probes on the GHA hosts :)
```
krisnova, line 186: TIL the ignore attribute :)

auraed/src/ebpf/loader.rs
JeroenSoeters: There is only one condition in which this will happen and this is when our user-space buffer is not large enough to hold all the events we read from the per-CPU buffer: https://github.com/JeroenSoeters/aurae/blob/sensible-buffer-sizes/auraed/src/ebpf/loader.rs#L156-L157

`.read_events()` will store in the buffer whatever fits and everything that does not fit is "dropped"

krisnova: My quick review is telling me we don’t need to be concerned about this condition in that case — should we panic in the event of a drop? Or at least increase the log level?

JeroenSoeters: Actually... rereading my own code :) this should never happen anymore as those buffers are now the same size as I pass the buffer size now also to `perf_array.open()`. Maybe we should make this an error instead of a warning? 🤔

krisnova (dismissed):

auraed/src/observe/mod.rs
JeroenSoeters: :)

auraed/src/ebpf/loader.rs
krisnova: Heh we had the same conclusion - yes let’s go with error level

JeroenSoeters: https://github.com/JeroenSoeters/aurae/blob/sensible-buffer-sizes/auraed/src/ebpf/loader.rs#L139-L140

JeroenSoeters: I've changed all the warnings to errors, as they all really shouldn't happen.

krisnova (approved):

PR #388, opened 2023-02-09
introduce cache for PodSandbox

### LINKED ISSUE #385: Implement Cache for CRI
Right now the CRI service code is blocked on adding a caching mechanism to the code base (similar to what we have [done with cells](https://github.com/aurae-runtime/aurae/blob/main/auraed/src/cells/cell_service/cell_service.rs)).

Can we please add the get/set cache code into the `cri_service.rs` file in place of the existing todo's?

For example in order for us to stop a pod sandbox:

```rust 
    async fn stop_pod_sandbox(
        &self,
        _request: Request<StopPodSandboxRequest>,
    ) -> Result<Response<StopPodSandboxResponse>, Status> {
        // TODO: Pull sandbox from cache
        // TODO: sandbox.kill()
        todo!()
    }
```

We need to first have the cached pod sandbox available, then we can call `sandbox.kill()` on the item retrieved from the cache.

dmah42:
Fixes #385


auraed/src/cri/error.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub(crate) type Result<T> = std::result::Result<T, RuntimeServiceError>;
+
+#[derive(Debug, Error)]
+pub enum RuntimeServiceError {
+    #[error("sandbox '{sandbox_id}' already exists")]
+    SandboxExists { sandbox_id: String },
+    #[error("sandbox '{sandbox_id}' not found")]
+    SandboxNotFound { sandbox_id: String },
+    #[error("Failed to kill sandbox '{sandbox_id}': {error}")]
```
future-highway, line 44: I know it's weird wording sometimes but I tend to like the error messages to start the same. "sandbox {sandbox_id}...". Only a preference

auraed/src/cri/runtime_service.rs
```diff
@@ -115,72 +125,111 @@ impl runtime_service_server::RuntimeService for RuntimeService {
...
-        let mut sandbox = sandbox_builder
-            .with_root_path(Path::new(AURAE_PODS_PATH).join(sandbox_id.clone()))
-            .expect("Setting pods directory")
-            .as_init(Path::new(AURAE_BUNDLE_PATH).join(AURAE_SELF_IDENTIFIER))
-            .with_systemd(false)
-            .build()
-            .expect("failed building pod sandbox: ensure valid OCI spec and proper container starting point");
-
-        sandbox.start().expect("starting pod sandbox");
+        let mut sandbox: Option<Container>;
```
future-highway, line 128: It looks like you are using option  to initialize to none temporarily. The block can return the sandbox. 'let sandbox = { ... };'

auraed/src/cri/runtime_service.rs
```diff
@@ -115,72 +125,111 @@ impl runtime_service_server::RuntimeService for RuntimeService {
...
+        request: Request<StopPodSandboxRequest>,
     ) -> Result<Response<StopPodSandboxResponse>, Status> {
-        // TODO: Pull sandbox from cache
-        // TODO: sandbox.kill()
-        todo!()
+        let sandbox_id = request.into_inner().pod_sandbox_id;
+
+        let mut sandboxes = self.sandboxes.lock().await;
+        let sandbox = sandboxes.get_mut(&sandbox_id)?;
+        sandbox.kill(Signal::SIGKILL, false).map_err(|e| {
```
future-highway, line 180: Should we not be removing it from the cache after the kill?

auraed/src/cri/runtime_service.rs
```diff
@@ -115,72 +125,111 @@ impl runtime_service_server::RuntimeService for RuntimeService {
...
     async fn remove_pod_sandbox(
         &self,
-        _request: Request<RemovePodSandboxRequest>,
+        request: Request<RemovePodSandboxRequest>,
     ) -> Result<Response<RemovePodSandboxResponse>, Status> {
-        // TODO: Delete sandbox from cache
-        // TODO: Ensure /var/run/aurae/pods/$container_name is destroyed
-        todo!()
+        let mut sandboxes = self.sandboxes.lock().await;
+        sandboxes.remove(&request.into_inner().pod_sandbox_id)?;
```
future-highway, line 191: What if it wasn't killed?

auraed/src/cri/sandbox_cache.rs
```diff
@@ -0,0 +1,83 @@
...
+            };
+        Ok(sandbox)
+    }
+
+    pub fn list(&self) -> Result<Vec<&Container>> {
+        Ok(self.cache.values().collect())
+    }
+
+    pub fn remove(&mut self, sandbox_id: &String) -> Result<()> {
+        if !self.cache.contains_key(sandbox_id) {
```
future-highway, line 75: You can check the return of remove. If None, then error

auraed/src/cri/sandbox_cache.rs
```diff
@@ -0,0 +1,83 @@
...
+        sandbox: Container,
+    ) -> Result<()> {
+        if self.cache.contains_key(&sandbox_id) {
+            return Err(RuntimeServiceError::SandboxExists { sandbox_id });
+        }
+        let _ = self.cache.insert(sandbox_id, sandbox);
+        Ok(())
+    }
+
+    pub fn get_mut(&mut self, sandbox_id: &String) -> Result<&mut Container> {
```
future-highway, line 56: We should eventually refactor to have most logic in the cache if possible and then not have a mut ref exposed.

auraed/src/cri/runtime_service.rs
```diff
@@ -115,72 +125,111 @@ impl runtime_service_server::RuntimeService for RuntimeService {
...
+                .start()
+                .expect("starting pod sandbox");
+        }
 
-        // TODO: Cache sandbox
+        // NOTE: we're creating a libcontainer::container and calling it `sandbox`,
+        // so that's what goes in the cache.  but i think the rest of this service
+        // differentiates between the outer "pod" and the inner "container"s so we
+        // might want to do the same.
+        let mut sandboxes = self.sandboxes.lock().await;
```
future-highway, line 166: Since the lock is taken after the sandbox is created, do we have a race condition. The cache will correctly error on the second one, but then we need to clean up.

auraed/src/cri/runtime_service.rs
dmah42: ack

auraed/src/cri/runtime_service.rs
dmah42: i thought about that but there's an explicit `remove_pod_sandbox` call just below, so i thought that might be the equivalent of "stop" and "free".

auraed/src/cri/runtime_service.rs
dmah42: ack

auraed/src/cri/sandbox_cache.rs
dmah42: ack

auraed/src/cri/runtime_service.rs
dmah42: ack.. moved the lock up.

auraed/src/cri/runtime_service.rs
```diff
@@ -115,72 +128,115 @@ impl runtime_service_server::RuntimeService for RuntimeService {
...
+                .start()
+                .expect("starting pod sandbox");
+            sandbox
+        };
+
+        // NOTE: we're creating a libcontainer::container and calling it `sandbox`,
+        // so that's what goes in the cache.  but i think the rest of this service
+        // differentiates between the outer "pod" and the inner "container"s so we
+        // might want to do the same.
+        sandboxes.add(sandbox_id.clone(), sandbox.expect("a sandbox"))?;
```
krisnova, line 169: So I think we want a new `Sandbox` struct that contains a set of containers inside 🙁 

I think the members of the struct should be:
 
 - Init container (in this case the freshly spawned auraed)
 - A []set of tenant containers that sit along side the init container
 - Probably more in the future?

krisnova (approved):

krisnova (approved):

PR #389, opened 2023-02-10
VM Service

mccormickt:
Defines the VM service in protobuf. Inspired by features from the [firecracker-containerd ](https://github.com/firecracker-microvm/firecracker-containerd/blob/main/proto/firecracker.proto) API as well as the VmResources struct used in the VMM crate in [Firecracker](https://github.com/firecracker-microvm/firecracker/blob/8ce7a756cd30a605d3e27753bc5e6c26fc58c7f5/src/vmm/src/resources.rs#L101-L124). 

This should give a good starting point, however lots of configuration options are exposed in this API. We could afford to be more opinionated and reduce the amount of VM configuration available initially if we so choose.


api/v0/virtual_machines/virtual_machines.proto
```diff
@@ -0,0 +1,178 @@
...
+  /// being mounted as read-write. Defaults to false, in which case the
+  /// rootfs is mounted as read-only.
+  bool is_writeable = 3;
+
+  /// (Optional) CacheType specifies the caching strategy for the block device.
+  /// The supported caching strategies are: "Unsafe"(default) and "Writeback".
+  string cache_type = 5;
+}
+
+message DriveMount {
```
jan0ski, line 151: For drives, I think we could trim out `cache_type`, `part_uuid`, `fs_type`, and `options` for now to keep it simple for the initial implementation

api/v0/virtual_machines/virtual_machines.proto
```diff
@@ -0,0 +1,178 @@
...
+  /// the VM.
+  string primary_addr = 1;
+
+  /// GatewayAddr specifies the default gateway that a network interface
+  /// should use inside the VM.
+  string gateway_addr = 2;
+
+  /// Nameservers is a list of nameservers that the VM will be configured
+  /// to use internally. Currently only up to 2 nameservers can be specified
+  /// (any more in the list will be ignored) and configuration is provided
```
jan0ski, line 125: Will want to remove much of the opinionated pieces on the implementation, until we decide what our defaults are.

api/v0/virtual_machines/virtual_machines.proto
```diff
@@ -0,0 +1,155 @@
...
+ *   distributed under the License is distributed on an "AS IS" BASIS,        *
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ *   See the License for the specific language governing permissions and      *
+ *   limitations under the License.                                           *
+ *                                                                            *
+\* -------------------------------------------------------------------------- */
+
+syntax = "proto3";
+
+package aurae.vms.v0;
```
dmah42, line 33: if the package is `.vms` then the file should probably be `vms/vms.proto` too i think.

dmah42:
it's a good starting point.. how much do you want to implement before merging?

api/v0/virtual_machines/virtual_machines.proto
jan0ski: I'm thinking of avoiding `vm` as it might be a little too brief for package names (and is quite overloaded in firecracker libraries). I'll change this to `virtual_machines` for clarity.

jan0ski:
> it's a good starting point.. how much do you want to implement before merging?


> it's a good starting point.. how much do you want to implement before merging?

api/v0/virtual_machines/virtual_machines.proto
```diff
@@ -0,0 +1,155 @@
...
+}
+message VmServiceStopResponse{}
+
+
+/// An Aurae virtual machine
+message VirtualMachine {
+  // The identifier of the VM
+  string id = 1;
+  
+  /// The memory size of VM
```
dmah42, line 81: i think these should be "//" not "///". it's proto, not rustdoc.  sorry for the nit.

api/v0/virtual_machines/virtual_machines.proto
```diff
@@ -0,0 +1,155 @@
...
+  uint32 mem_size_mb = 2;
+  
+  /// The number of vCPUs for the VM
+  uint32 vcpu_count = 3;
+  
+  /// The path to the VM kernel image
+  string kernel_img_path = 4;
+  
+  /// Arguments to pass to the kernel
+  string kernel_args = 5;
```
dmah42, line 91: should this be repeated for multiple args instead of assuming a split character?

dmah42:
dropping this in as a framework sounds great so we can build on it.  have you seen the Validation stuff? check the other services for how that works.  that might be a good next step given how many fields are strings but have semantic expectations.

dmah42:
`api/v0/virtual_machines/virtual_machines.proto: File does not reside within any path specified using --proto_path (or -I).  You must specify a --proto_path which encompasses this file.  Note that the proto_path must be an exact prefix of the .proto file names -- protoc is too dumb to figure out when two paths (e.g. absolute and relative) are equivalent (it's harder than you think).`

you might want some Makefile changes.

api/v0/virtual_machines/virtual_machines.proto
jan0ski: I was definitely rustdoc-ing here, thanks!

api/v0/virtual_machines/virtual_machines.proto
jan0ski: Happy to break this up if that's more intuitive.  I've only seen kernel args as one concatenated string but repeating is likely better UX and easier to validate.

krisnova (approved):

Issue #365, opened 2023-02-05
Integration tests for auraescript examples


JeroenSoeters:
It would be nice if we had a test harness around the auraescript examples we ship. What if we would add comments with the expected output in the example TypeScript files and we'd write a test runner that would parse all those comments, builds an array of expected JSON output and compares that against the actual output of the script. Like this:

```
let cells = new runtime.CellServiceClient();
const cellName = "ae-sleeper-cell";
const nestedCellName = "ae-sleeper-cell/nested-sleeper"

// [ Allocate ]
let allocated = await cells.allocate(<runtime.CellServiceAllocateRequest>{
    cell: runtime.Cell.fromPartial({
        name: cellName,
        cpu: runtime.CpuController.fromPartial({
            weight: 2, // Percentage of CPUs
            max: 400 * (10 ** 3), // 0.4 seconds in microseconds
        }),
    })
});
helpers.print(allocated)

// output
// {
//      "cellName": "ae-sleepercell",
//      "v2": true
// }

// [ Start ]
let started = await cells.start(<runtime.CellServiceStartRequest>{
    cellName,
    executable: runtime.Executable.fromPartial({
        command: "/usr/bin/sleep 42",
        description: "Sleep for 42 seconds",
        name: "sleep-42"
    })
})
helpers.print(started)

//  output
//  {
//    "pid": "<pid>"
//  }
```

With `"key" : " value"` we mean assert that the key exists in the output and assert that the values are equal. With `"key" : "<value>"` we indicate that we expect the key to be present and has _any_ value. We could create a `...` pattern for streams when we support them:

```
// output
// [
//     {
//        "some": "stream"
//     },
//     ...   
// ] 
```

dmah42:
i'm not convinced we need to include the expected output in the script.  what i would expect is a framework that allows us to:
1. start the auraed daemon
2. run a typescript script, capturing the output
3. check the output against expectations

given how broadly-used typescript is, i'm a bit surprised there isn't a framework already for this..

JeroenSoeters:
I should probably have been more clear in that what I am proposing is exactly that :) It's just a matter of where we want the expectations to live. Here I'm inlining those expectations with the example scripts, which has the advantage that you only have to edit a single file which is your example and your test. We could automatically run the framework against every file in the examples folder. Also, if you're just browsing the repo, you don't necessarily have to run the examples to see what output they produce.

[...]

PR #391, opened 2023-02-10
Introduce the ability to specify workload type for `GetPosixStreamSignals`

JeroenSoeters:
This PR allows for specifying a workload when retrieving the POSIX signals stream. Right now only 1 type of workload is supported (`Cell`) and only a single `Cell` can be specified. 

To achieve this the eBPF probe now surfaces the `cgroup_id` and the `observe_service` has a cache that allows for looking up the `cgroup_path` for a `cgroup_id`. With the `cgroup_path` we can filter out signals relevant to a specific cell.


api/v0/observe/observe.proto
```diff
@@ -52,7 +52,30 @@ service ObserveService {
   rpc GetPosixSignalsStream(GetPosixSignalsStreamRequest) returns (stream GetPosixSignalsStreamResponse) {}
 }
 
-message GetPosixSignalsStreamRequest { 
+/// Request a stream of POSIX signals
+message GetPosixSignalsStreamRequest {
+  /// The scope of this request. If no scope is specified, a stream of all POXIS
+  /// signals on the host will be returned.
+  oneof workload {
```
dmah42, line 59: this might generate nicer code as an enum and an opaque string "id".  oneof is pretty nasty, especially in some languages.

dmah42, line 59: also, should this be repeated?

e2e-tests/src/observe/get_posix_signals_stream.rs
```diff
@@ -0,0 +1,102 @@
+#[cfg(test)]
+mod tests {
+    use std::{sync::Arc, time::Duration};
```
dmah42, line 3: e2e tests are cool, but do we even have unit tests for this?

JeroenSoeters: See the discussion here: https://github.com/aurae-runtime/aurae/issues/365

We have a unit test in the `observe` service, but once we are going to test observing specific workloads we're going to need to do a lot of gymnastics on that unit test.

api/v0/observe/observe.proto
JeroenSoeters: We could introduce a "Workload" message with an enum and a string "id" and have this repeated. I need to think this thru a bit though what this means in terms of implementation details for this specific endpoint, as it could get pretty tricky (and not very intuitive) to aggregate POSIX signals for a mix of cells, containers, vms etc.

api/v0/observe/observe.proto
```diff
@@ -52,7 +52,30 @@ service ObserveService {
...
+    Container container = 3;
+    VirtualMachine vm = 4;
+  }
+}
+
+message Host {
+}
+
+message Cell {
+  string cell_name = 1;
```
future-highway, line 70: Are we going to have more fields? Wondering why these are wrapped in messages?

e2e-tests/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "e2e-tests"
```
future-highway, line 2: Is there a reason for a dedicated crate over a test directory?

e2e-tests/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+aurae-client = { workspace = true }
+aurae-proto = { workspace = true }
+futures-util = "0.3.26"
+tokio = { workspace = true }
+uuid = { version = "1.2.2", features = ["v4"] }
```
future-highway, line 13: I think we use this dep in multiple crates. Let's move it to the workspace level?

e2e-tests/Cargo.toml
future-highway: We can iterate later

e2e-tests/Cargo.toml
JeroenSoeters: `futures-util`, `uuid` or both? :)

api/v0/observe/observe.proto
JeroenSoeters: What do you think about how the API looks now @dmah42? I will create a separate issue to discuss observing multiple workloads.

api/v0/observe/observe.proto
JeroenSoeters: I've changed it now as per @dmah42's suggestion. These were mapped in messages b/c it was a `oneof`. Maybe I was thinking too much about how this would be mapped to Rust code though instead of what would be more idiomatic protobuf :)

api/v0/observe/observe.proto
JeroenSoeters: Also, if we want to avoid using `oneof` b/c of how this gets compiled in different languages, should we capture this in the stdlib docs as a "general rule"? https://aurae.io/stdlib/

e2e-tests/Cargo.toml
future-highway: I was thinking uuid, but both it they are used in multiple crates. I've been "reviewing" from my phone for the last week, so it's hard to keep it all in my head and get a full picture of the changes.

api/v0/observe/observe.proto
dmah42: we should probably point to https://google.aip.dev/ and specifically https://google.aip.dev/search?q=oneof which doesn't say avoid it, but it does provide some trade-offs.

api/v0/observe/observe.proto
```diff
@@ -52,7 +52,22 @@ service ObserveService {
   rpc GetPosixSignalsStream(GetPosixSignalsStreamRequest) returns (stream GetPosixSignalsStreamResponse) {}
 }
 
-message GetPosixSignalsStreamRequest { 
+/// Request a stream of POSIX signals
+message GetPosixSignalsStreamRequest {
+  /// The workload to which te response will be scoped. If no workload is
+  /// specified, a stream of all POSIX signals on the host will be returned.
+  Workload workload = 1;
```
dmah42, line 59: i personally prefer this (with notes) but i think it should be repeated.

api/v0/observe/observe.proto
```diff
@@ -52,7 +52,22 @@ service ObserveService {
...
+}
+
+enum WorkloadType {
+  WORKLOAD_CELL = 0;
```
dmah42, line 63: value 0 should be WORKLOAD_TYPE_UNSPECIFIED

api/v0/observe/observe.proto
```diff
@@ -52,7 +52,22 @@ service ObserveService {
...
+  WORKLOAD_CONTAINER = 1;
```
dmah42, line 64: these should be WORKLOAD_TYPE_CONTAINER etc

auraed/src/observe/cgroup_cache.rs
```diff
@@ -0,0 +1,130 @@
...
+        if let Some(path) = self.cache.get(&ino) {
+            Some(path.clone())
+        } else {
+            self.refresh_cache();
+            self.cache.get(&ino).cloned()
+        }
+    }
+
+    fn refresh_cache(&mut self) {
+        fs::read_dir(&self.root)
```
future-highway, line 57: We have nested cgroups, does this read those directories too? Docs don't make it seem like it. I think we already have a dep on https://crates.io/crates/walkdir which may be helpful

auraed/src/observe/cgroup_cache.rs
```diff
@@ -0,0 +1,130 @@
...
+    cache: HashMap<u64, OsString>,
+}
+
+impl CgroupCache {
+    pub fn new(root: OsString) -> Self {
+        Self { root, cache: HashMap::new() }
+    }
+
+    pub fn get(&mut self, ino: u64) -> Option<OsString> {
+        if let Some(path) = self.cache.get(&ino) {
```
future-highway, line 48: Do we have to be concerned that we will cache the ino, then the system will reuse it and our cache will point to the wrong path?

auraed/src/observe/cgroup_cache.rs
```diff
@@ -0,0 +1,130 @@
...
+\* -------------------------------------------------------------------------- */
+use std::collections::HashMap;
+use std::ffi::OsString;
+use std::fs;
+use std::os::unix::prelude::DirEntryExt;
+use tracing::warn;
+
+/// Used for looking up cgroup paths by inode number
+struct CgroupCache {
+    root: OsString,
```
future-highway, line 38: Is this ever not sys/fs/cgroups?

JeroenSoeters: Only in my tests :)

auraed/src/observe/cgroup_cache.rs
JeroenSoeters: Don't really want to do this now, but at some point, it might be worth investigating how Linux itself does inode cache invalidation: https://github.com/torvalds/linux/blob/master/fs/inode.c

future-highway: Maybe put a todo so that we know to revisit it

JeroenSoeters: I'll add this to the LRU TODO that's already in this file (and might or might not actually solve this, need to think this thru a bit)

api/v0/observe/observe.proto
JeroenSoeters: Should we iterate on the API later? Making this `repeated` warrants some discussion imo. Like how would the response look like in these cases? For a single workload, the response is just a stream of signals. If you would be able to specify multiple cells/containers etc we would need to return tuples (id, signal). If we allow for a mix of workload types we need to return 3-tuples (workload_type, id, signal) for the user to make sense of the response. Maybe something for an RFC?

e2e-tests/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "e2e-tests"
+version = "0.1.0"
+edition = "2021"
+
```
future-highway, line 5: Can you add license and authors fields, and set version to 0.0.0

e2e-tests/src/lib.rs
```diff
@@ -0,0 +1 @@
+mod observe;
```
future-highway, line 1: Setup lints

```rust
// Lint groups: https://doc.rust-lang.org/rustc/lints/groups.html
#![warn(future_incompatible, nonstandard_style, unused)]
#![warn(
    improper_ctypes,
    non_shorthand_field_patterns,
    no_mangle_generic_items,
    unconditional_recursion,
    unused_comparisons,
    while_true
)]
#![warn(
    missing_debug_implementations,
    missing_docs,
    trivial_casts,
    trivial_numeric_casts,
    unused_extern_crates,
    unused_import_braces,
    unused_results
)]
#![warn(clippy::unwrap_used)]
```

ebpf-shared/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 #[repr(C)]
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Signal {
+    pub cgroupid: u64,
```
future-highway, line 6: `cgroup_id`?

ebpf-shared/src/lib.rs
```diff
@@ -3,6 +3,7 @@
...
     pub signr: i32,
```
future-highway, line 7: `signal`?

future-highway (commented):
Just leaving comments. Didn't do a full review.

ebpf-shared/src/lib.rs
JeroenSoeters: Sorry for some reason I always go minimal when writing C :)

ebpf-shared/src/lib.rs
JeroenSoeters: I'll actually call this `signum`, I would like with those "shared" types to align as closely as possible with kernel code to be explicit what the fields are.

api/v0/observe/observe.proto
```diff
@@ -52,7 +52,23 @@ service ObserveService {
...
+message GetPosixSignalsStreamRequest {
+  /// The workload to which te response will be scoped. If no workload is
+  /// specified, a stream of all POSIX signals on the host will be returned.
+  Workload workload = 1;
+}
+
+enum WorkloadType {
+  WORKLOAD_TYPE_UNSPECIFIED = 0;
+  WORKLOAD_TYPE_CELL = 1;
+  WORKLOAD_TYPE_CONTAINER = 2;
```
krisnova, line 65: This should be a PodSandbox and not a container to keep with the "service level" types.

Also we will likely want a container type in the future, but we can cross that bridge later.

ebpf-shared/src/lib.rs
future-highway: I agree. If the field name is being taken from the kernel, stick to that name (we try to do the same with cgroups).

krisnova (approved):

PR #395, opened 2023-02-12
Re-enable cargo udeps and remove unused dependencies

### LINKED ISSUE #306: Re-enable check udeps
I understand that `cargo udeps` requires `+nightly` which introduces some compile-time problems, specifically when compiling with musl.

We need to figure out a udeps strategy and re-enable the build at some point. I turned it off in https://github.com/aurae-runtime/aurae/pull/305

The commit: https://github.com/aurae-runtime/aurae/pull/305/commits/10ebaf05a525c0d626269378831dfe1fcd4c2aa1

tlaferriere:
I moved `simplelog` and `multi-log` crates to `dev-dependencies` because they were only used in unittests. As for running udeps as nightly and building for musl, ~it works on my machine :shrug:. We will see when the gh actions run~ the gh actions passed, maybe it was just a buggy version of the nightly musl target or the cargo-udeps crate.

Closes #306

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -24,7 +24,7 @@ version = "0.0.0"
 dependencies = [
  "aer-macros",
  "anyhow",
- "clap 4.1.4",
+ "clap",
  "client",
  "futures-util",
  "proto",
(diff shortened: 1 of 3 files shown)


dmah42 (approved):

dmah42:
thank you!

PR #404, opened 2023-02-13
Introduce auraed integration testing

future-highway:
I had to refactor lib.rs and main.rs to get this set up. In the process though, we are now respecting the runtime settings everywhere (I think), but I did do it by introducing a global variable.


JeroenSoeters:
I was actually planning on getting rid of that test in `cell_service.ts` and move it to the e2e test crate (which is not yet merged). This would address https://github.com/aurae-runtime/aurae/issues/365. Should we come up with some heuristics for when to write an integration test vs an e2e test?

future-highway:
As I'm sure you've seen, many of our tests are not actually unit tests, though we have only been using rust's built in support for unit test. So, I just wanted a single test to use to set this up; it is not an opinion of where the test should live.

That being said, my immediate thought is that I think we should try to use rust's built in support for integration tests where possible and use e2e tests if needed. I can probably be easily persuaded.

JeroenSoeters:
That's fine. Integration tests where possible and e2e where needed is what I'm thinking as well. With an extra note that we'll aim to have at least one (and ideally not many more) e2e test per gRPC endpoint.

Because you picked this specific test, I wanted to make sure we're not going in a direction where we have one e2e test per endpoint and then another integration test, testing everything but the wrapping and unwrapping of those tonic gRPC requests/responses. Otherwise, all those tests would be testing 95% the same code paths.

future-highway:
@JeroenSoeters We should also keep in mind, rust will run all tests in a crate in parallel (I noticed you set threads to 1 to prevent this in the other PR), and each file under the tests dir is considered a crate. Tests in different crates are run consecutively. In all scenarios we will have the system as shared state.

For now, I think it's better for you to proceed with your PR as if this one doesn't exist and, assuming this PR is accepted and merged, we will iterate the testing setup.

dmah42 (commented):
I think it's a shame we are losing the idiomatic rust pattern of tests in the same file as the code under test.

I'm also extremely wary of prioritising integration tests over unit tests (this and the statement above are linked).

is the intention to still have unit tests alongside the code with integration tests where necessary, and we don't have unit tests, or to eschew unit tests in favour of integration and e2e tests?

the latter would be a significant mistake for long term maintenance of the project. the former is fine but we really should get some effort into creating unit tests.

this may mean refactoring code to make it easier to unit test but that's also ideal for long term maintenance.

tl;dr: if the only way we can test our code is integration and e2e tests, we need to change the code.

future-highway:
So the first step I think I should take is to try breaking this PR in 2 so we can review the refactoring on its own. I'm still thinking about the rest of your feedback

future-highway:
> I think it's a shame we are losing the idiomatic rust pattern of tests in the same file as the code under test.

I agree it's nice to have tests near what is being tested, and I also like that rust allows that. I wouldn't say integration tests are not idiomatic though, as they are also part of cargo.

> I'm also extremely wary of prioritising integration tests over unit tests (this and the statement above are linked).
> 
> is the intention to still have unit tests alongside the code with integration tests where necessary, and we don't have unit tests, or to eschew unit tests in favour of integration and e2e tests?

I think we should continue to have unit tests where possible, but some of our tests that are run as unit tests are not actual unit tests. For example, the ones in cells.rs and cell.rs. I don't think we have to have a rule that those tests need to be actual unit tests and, at least, `test_cant_unfree` in cell.rs looks to be impossible to test from anywhere else as the code is currently written. However, the tests in cells.rs are all testing things that should also be covered by integration tests which will make them redundant.

> the latter would be a significant mistake for long term maintenance of the project. the former is fine but we really should get some effort into creating unit tests.
> 
> this may mean refactoring code to make it easier to unit test but that's also ideal for long term maintenance.
> 
> tl;dr: if the only way we can test our code is integration and e2e tests, we need to change the code.

We would have to do some refactoring to allow for testing smaller parts of the code.

future-highway:
Marked this as a draft again until we decide how we want to proceed.

JeroenSoeters:
Agreeing with @future-highway here. I believe we should write unit tests where possible and integration/e2e tests where needed. In the past, when I've written code like this, where you have a relatively thin layer that mostly interfaces with the kernel, I typically ended up with more integration tests than unit tests just because the code didn't lend itself very well to unit testing. Happy to look into potential opportunities for refactoring though.

dmah42:
> > I think it's a shame we are losing the idiomatic rust pattern of tests in the same file as the code under test.
> 
> I agree it's nice to have tests near what is being tested, and I also like that rust allows that. I wouldn't say integration tests are not idiomatic though, as they are also part of cargo.

fair.  i have a long history of problems with integration and e2e tests and a strong preference to avoid them whenever possible.  in terms of long-term maintenance of a code-base, they end up being slow (limiting local dev velocity), flaky (causing churn and requiring more attention), and often fail not due to regressions but due to refactorings of the underlying code that change some aspect of the setup/construction of the system under test.

> 
> > I'm also extremely wary of prioritising integration tests over unit tests (this and the statement above are linked).
> > is the intention to still have unit tests alongside the code with integration tests where necessary, and we don't have unit tests, or to eschew unit tests in favour of integration and e2e tests?
> 
> I think we should continue to have unit tests where possible, but some of our tests that are run as unit tests are not actual unit tests. For example, the ones in cells.rs and cell.rs. I don't think we have to have a rule that those tests need to be actual unit tests and, at least, `test_cant_unfree` in cell.rs looks to be impossible to test from anywhere else as the code is currently written. However, the tests in cells.rs are all testing things that should also be covered by integration tests which will make them redundant.

this is the thing i disagree with the most.  integration tests do not make unit tests redundant.  the unit tests allow you to focus in on any regressions that integration tests highlight, mitigating some of the difficulties with integration test maintenance.  we absolutely need to have as much coverage from unit tests as possible even if integration tests cover the same code.  integration tests should be testing specifically the interaction between systems, each of which has maximal unit test coverage.  this is how we build for the long-term.

> 
> > the latter would be a significant mistake for long term maintenance of the project. the former is fine but we really should get some effort into creating unit tests.
> > this may mean refactoring code to make it easier to unit test but that's also ideal for long term maintenance.
> > tl;dr: if the only way we can test our code is integration and e2e tests, we need to change the code.
> 
> We would have to do some refactoring to allow for testing smaller parts of the code.

i think this is worthwhile if we want to have a project we can continue to maintain.  having a balance between "sensible public APIs" and "testable units" is difficult, but this is what makes software engineering fun.  we shouldn't refactor _just_ to make things testable if it breaks APIs, but we should make sure any public API for a class (struct i guess) is testable as much as possible.

what i don't know about (idiomatic) rust is how things like dependency injection are handled.  for example, fakes and mocks in things like Java or Go.  I think we might need something like that at some point.

the short(ish) version: i think it's fine and probably useful to introduce integration testing to cover interactions between systems under test.  i don't think we should do that at the expense of unit tests that cover each of those systems _as much as possible_. we should also write the code in a way that makes it more possible to cover the code with unit tests.

apologies: you've hit on a nerve here for me as i've spent a huge amount of my career managing this stuff within teams in the $dayjob and seen first-hand the benefits of getting this right early (and the costs of getting it wrong).

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -421,97 +421,3 @@ impl cell_service_server::CellService for CellService {
...
-    use crate::cells::cell_service::validation::{
-        ValidatedCell, ValidatedCpuController, ValidatedCpusetController,
-        ValidatedMemoryController,
-    };
-    use crate::{AuraedRuntime, AURAED_RUNTIME};
-    use iter_tools::Itertools;
-    use test_helpers::*;
-
-    #[tokio::test]
-    async fn test_list() {
```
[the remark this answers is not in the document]
future-highway: Note: It's very possible my thought on what is a unit test is wrong.

For this test in particular, it creates actual cgroups on the node, which to me rules it out as a unit test.

auraed/tests/common/cells.rs
```diff
@@ -82,12 +79,12 @@ impl ExecutableBuilder {
     pub fn new() -> Self {
         Self {
             name: format!("ae-sleeper-{}", uuid::Uuid::new_v4()),
-            command: "sleep 400".to_string(),
+            command: "sleep 5".to_string(),
```
JeroenSoeters, line 82: The reason why I prefer a longer timeout is so I can poke around in `/proc` if I'm trying to figure something out.

auraed/tests/cells_list.rs
```diff
@@ -0,0 +1,113 @@
+use client::cells::cell_service::CellServiceClient;
+use common::cells::CellServiceAllocateRequestBuilder;
+use iter_tools::Itertools;
+use proto::cells::{CellGraphNode, CellServiceListRequest};
+use test_helpers::*;
+
+mod common;
+
+#[test_helpers_macros::shared_runtime_test]
+async fn cells_list_must_list_allocated_cells_recursively() {
```
JeroenSoeters, line 10: Did you accidentally put this test here instead of the test that was initially in `e2e-tests/src/cells_list.rs`? IMO we should remove this test as it doesn't really belong here (if we consider the tests in this folder e2e tests). Also, everything this test is testing (and more) was already covered in the above-mentioned e2e test.

FWIW I am introducing a lot more unit-style testing in my upcoming PR to move us to a healthier pyramid shape

Cargo.toml
```diff
@@ -62,11 +61,12 @@ serde = "1.0"
 serde_json = "1.0"
 syn = { version = "1.0", features = ["full"] } # used in macros, so full doesn't affect binary size
 test-helpers = { path = "./crates/test-helpers" }
+test-helpers-macros = { path = "./crates/test-helpers-macros" }
```
JeroenSoeters, line 64: Why do we have 2 crates, `test-helpers` and `test-helper-macros` while both crates only contain macros? :)

future-highway: proc-macros need their own crate and have `proc-macro = true` in cargo.toml.

auraed/tests/cells_list.rs
future-highway: I'll revert

auraed/tests/cells_list.rs
```diff
@@ -0,0 +1,199 @@
...
+            },
+        ],
+    };
+
+    // Assert that the actual response matches the expected
+    assert_eq!(list_response, expected);
+}
+
+#[test_helpers_macros::shared_runtime_test]
+async fn cells_list_must_list_deeply_nested_cells_recursively() {
```
JeroenSoeters, line 96: Are we ok deleting this? I'm not sure this adds a lot of value. IMO it just makes the tests slower and potentially confuses future contributors.

auraed/tests/cells_list.rs
```diff
@@ -0,0 +1,199 @@
+use client::cells::cell_service::CellServiceClient;
```
JeroenSoeters, line 1: nitpick: the service is called `cell_service`, so maybe we should call the test `cell_list`?

auraed/tests/cells_list.rs
future-highway: I wanted to specifically test a triple nest (I know we have the unit test too). Also, I don't love that the e2e test uses builders that hide most of the fields and then tests all the fields. Let's keep both for now.

PR #405, opened 2023-02-14
Settings refactoring

future-highway:
Extracted from #404. 

- Refactored lib.rs and main.rs. 
- We are should now be respecting the runtime settings everywhere, but I did do it by introducing a global variable.


auraed/src/logging/stream_logger.rs
```diff
@@ -41,6 +41,7 @@ pub struct StreamLogger {
 }
 
 impl StreamLogger {
+    #[allow(dead_code)]
```
dmah42, line 44: can we "allow(unused)" instead? unused suggests "i'm going to use this later" .. dead code suggests "this is no longer needed but i'm scared to delete it".

dmah42 (approved):

PR #406, opened 2023-02-14
Add CPU quota burning example

jcorbin:
This is a draft / WIP PR for code sharing / reference in an issue ; please do not review

### DIFF
--- a/examples/burn.go
+++ b/examples/burn.go
@@ -1,0 +1,197 @@
+// This program bursn as much CPU as it can, while measuring runtime lag.
+//
+// CPU burn is achieved by computing random collatz trajectories.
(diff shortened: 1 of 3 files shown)


krisnova:
Hello 👋🏻 this looks like a very exciting example! Would you be interested in finishing this work so we can merge this in? I believe the community would love more concrete examples with Go like this! 

No rush, please keep the open PR as long as you need!

dmah42:
hi @jcorbin .. i know it's been a while since you opened this up, but it would be great to continue to build this out to a full example.  having something we can run and maintain as a real example, and see where there are gaps in our API, would be extremely useful.

PR #408, opened 2023-02-14
Migrate list test for cell service to an e2e test

JeroenSoeters:
Moved this test out of the `cell_service` into the e2e tests package, so we test the mapping to the tonic request/response objects as well. 

It is, however ignored for now as we need a fresh `auraed` instance with this way of matching the response. @future-highway maybe it is time to combine these with the macros you wrote for the integration tests?


e2e-tests/src/common/helpers.rs
```diff
@@ -31,6 +32,12 @@ pub async fn stop_in_cell(client: &AuraeClient, req: CellServiceStopRequest) {
     assert!(res.is_ok());
 }
 
+pub async fn list_cells(client: &AuraeClient) -> CellServiceListResponse {
+    let res = client.list(CellServiceListRequest {}).await;
+    assert!(res.is_ok());
+    res.expect("CellServiceListResponse").into_inner()
```
future-highway, line 38: I'm also not a fan of helpers for tests and think they should be used with caution. When something errors, the message will point to a line in the helper which may make it harder to understand where exactly a test failed (i.e., was it the first call to the helper, or was was it the 2nd?).

To simplify the code, using `.unwrap()` is allowed in tests, which should make for less of a want for the helpers.

e2e-tests/src/cells/list.rs
```diff
@@ -0,0 +1,88 @@
...
+                        children: vec![],
+                    }],
+                },
+            ],
+        };
+
+        // Assert that the actual response matches the expected
+        assert_eq!(
+            list_response, expected,
+            "got {:#?}\nexpected {:#?}",
```
future-highway, line 84: I think this will be similar to the default message (though that uses "left" and "right" if I remember correctly).

e2e-tests/src/cells/list.rs
```diff
@@ -0,0 +1,88 @@
...
+    use client::Client as AuraeClient;
+    use proto::cells::{Cell, CellGraphNode, CellServiceListResponse};
+
+    use crate::common::{
+        helpers::{allocate_cell, list_cells},
+        request_builders::CellServiceAllocateRequestBuilder,
+    };
+
+    #[tokio::test]
+    #[ignore = "this test requires a fresh Aurae daemon which right now we cannot guarantee"]
```
future-highway, line 12: Yes, we can probably use some of the  code from the other PR to deal with this. dma brought up some concerns though in the other PR, and I haven't formulated a response yet.

Re a "fresh" auraed, that isn't hard to get, but we will need to keep in mind whether the system is fresh or not and if that effects what we are testing. Cells should have a fresh system (best effort to clean up on `drop`), but I haven't looked into the other services yet.

e2e-tests/src/cells/list.rs
JeroenSoeters: The default doesn't pretty-print that's why I added this.

e2e-tests/src/common/helpers.rs
```diff
@@ -3,7 +3,8 @@ use std::sync::Arc;
 use client::observe::observe_service::ObserveServiceClient;
 use client::{cells::cell_service::CellServiceClient, Client as AuraeClient};
```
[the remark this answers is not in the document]
JeroenSoeters: No issues, I will change this back to `Client`.

e2e-tests/src/common/helpers.rs
JeroenSoeters: I don't have any strong feelings one way or the other. I moved it out because every call had an `assert!`, an `expect()`, and an `into_inner()`, and I felt it was making the tests less readable, but I understand the concern of being able to debug those tests.

I can remove it. It would mean that all those calls to the client will look like this:
```
client
.allocate(...)
.await
.unwrap()
.into_inner()
.cell_name
```
This would be _without_ the extra assert I put in there in the hope of creating some better test locality.

future-highway: I'm not convinced it's worth having.

`.unwrap()` covers the assert and `.cell_name` will be removed the moment we have a test that wants to read a different field on the response.

It's also not immediately clear what the `String` returned by `allocate_cell` is since the client doesn't have the strong typing we have in the auraed crate. That can be somewhat alleviated via doc comments, but then I think we are starting to introduce more of a maintenance burden.

Maybe someone else will provide an opinion as well.

e2e-tests/src/cells/list.rs
future-highway: Haven't used it, but it has a lot of downloads: https://docs.rs/pretty_assertions/latest/pretty_assertions/

e2e-tests/src/common/helpers.rs
JeroenSoeters: I'll leave it until tomorrow for others to comment; otherwise, I will get rid of all helpers but that one that intercepts the signals stream, as there is a bunch of stuff happening there that I wouldn't necessarily want to expose in the test. It would be nice to make that helper generic so we can reuse that pattern for other streaming endpoints. I spent some time wrestling with the borrow checker trying to do that, but ultimately the borrow checker came out on top. :)

JeroenSoeters: ☝️ what I was running into is that I wanted a function `intercept_stream` that takes a function that takes a `Client` and returns a `Future<Response<ResultStream<T>>>` and returns an `Arc<Mutex<Vec<T>>>` with the intercepted `T`'s. Because I then also needed to pass in the actual client obvs I ran into the issue that I needed an `async` closure that takes a parameter, and that was not a stable feature iirc.

future-highway: That code could be: 
```rust
// this is in the test
let res = client.get_posix_signals_stream(req).await.unwrap().into_inner();
let signals = stream_response_to_end(res).await.unwrap();
```

```rust
// helper
pub async fn stream_response_to_end<T>(
    response: tonic::Streaming<T>,
) -> Result<Vec<T>, tonic::Status> {
    tokio_stream::StreamExt::collect(response).await
}
```

`tokio-stream = "0.1.11"` needs to be added to cargo.toml

future-highway: Actually, these functions wouldn't be the same since my helper would require the stream to end, which it doesn't look like it ever does (looking at the implementation for `get_posix_signals_stream`). Hmm

JeroenSoeters: Ah, I was trying to do it with a single function. With 2 functions, I can do it like this:

```
// in test
let signals_stream = client
    .get_posix_signals_stream(
        GetPosixSignalsStreamRequestBuilder::new()
            .cell_workload(cell1_name.clone())
            .build(),
    )
    .await
    .unwrap()
    .into_inner();
let intercepted_signals = intercept_stream(signals_stream).await;
```

```
// in helper
pub async fn intercept_stream<T: Send + 'static>(
    res: tonic::Streaming<T>,
) -> Arc<Mutex<Vec<T>>> {
    let mut stream = res;

    let intercepted = Arc::new(Mutex::new(Vec::new()));
    let intercepted_in_thread = intercepted.clone();

    let _ignored = tokio::spawn(async move {
        while let Some(res) = futures_util::StreamExt::next(&mut stream).await {
            let res = res.expect("stream closed");
            let mut guard = intercepted_in_thread.lock().await;
            guard.push(res);
        }
    });

    intercepted
}
```
but because I'm now working with a `Arc<Mutex<Vec<Result<T>>>>` where `T` isn't even the actual type I'm interested in (`Signal`) but the gRPC response message (`GetPosixSignalsStreamResponse`). Working with this in the test becomes a bit of a beast:
```
// in test
let guard = intercepted_signals
    .lock()
    .await
    .iter()
    .map(|r| r.signal.clone().unwrap())
    .collect::<Vec<_>>();
```

e2e-tests/src/cells/list.rs
JeroenSoeters: I pulled it in just for the e2e tests for now. This will make comparing those gRPC response payloads a lot easier.

e2e-tests/src/common/helpers.rs
future-highway: I think I would try making it work with a channel if I were writing it, but if that works, I see it as an improvement. Ideally, any helpers wouldn't panic so that the error message doesn't point to a location in the helper.

JeroenSoeters: I think I see what you're saying. I'll try to summarize it here to see if we all agree on the approach. So the `tonic::Streaming` wraps a channel already, I believe, but it would be better to rewrite this `intercept_streaming(res: tonic::Streaming<T>)` into something like `expect_message(res: Receiver<T>, predicate: fn (T) -> bool, timeout: Duration)` and something similar for the negative case. This would make the positive cases a lot faster as now we're _always_ waiting 500ms which will add up with many tests. If we find ourselves writing a lot of negative tests (which _have_ to poll the channel for the full duration), this might be a cue to look for opportunities for unit testing those code paths, in line with the discussion with @dmah42 on the other issue.

future-highway (approved):
This looks good so far. I think it's a good place to merge and iterate.

JeroenSoeters:
@future-highway I rebased main, and now it looks like it needs another approval.

future-highway (approved):

PR #410, opened 2023-02-15
streaming logs from sub processes

dmah42:


api/v0/observe/observe.proto
```diff
@@ -86,7 +86,7 @@ message GetAuraeDaemonLogStreamRequest {
 // TODO: not implemented
 message GetSubProcessStreamRequest {
   LogChannelType channel_type = 1;
-  int64 process_id = 2;
+  int32 process_id = 2;
```
future-highway, line 89: We have been inconsistent with the type and name for pids. Above, `Signal` we use `int64 process_id` and in cells.proto we use `int32 pid`. Do you want to make them all the same in this PR?

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -214,11 +234,25 @@ impl CellService {
...
-        let _exit_status = executables
+        let executable = executables
             .stop(&executable_name)
             .await
             .map_err(CellsServiceError::ExecutablesError)?;
 
+        let pid = executable
+            .pid()
+            .map_err(CellsServiceError::Io)?
+            .expect("pid")
```
future-highway, line 245: This will always panic since an `Executable` in the `Stopped` `ExecutableState` will always return `None`. You can get the `pid` prior to stopping it.

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -187,8 +194,21 @@ impl CellService {
             .expect("pid")
             .as_raw();
 
-        // TODO: either tell the [ObserveService] about this executable's log channels, or
-        // provide a way for the observe service to extract the log channels from here.
+        // Ensure the observe service is aware of the executable's logs.
+        self.observe_service
+            .register_sub_process_channel(
```
future-highway, line 199: If these register calls error, the client will receive an error, but the executable will be running. It doesn't look like they should actually error, so maybe not a concern.

auraed/src/observe/observe_service.rs
```diff
@@ -70,13 +75,57 @@ impl ObserveService {
...
+        let mut consumer_list = self.sub_process_consumer_list.lock().await;
+        if let Some(channels) = consumer_list.get_mut(&pid) {
+            if channels.remove(&channel_type).is_none() {
+                return Err(ObserveServiceError::ChannelNotRegistered {
+                    pid,
+                    channel_type,
+                });
+            }
+        } else {
+            return Err(ObserveServiceError::NoChannelsForPid { pid });
```
future-highway, line 126: Should we ignore these errors in the CellService? Does it make sense to return an error to the client when the executable has been stopped and the channels are not registered?

future-highway (changes requested):
Block is for the pid panic comment

api/v0/observe/observe.proto
dmah42: i can.. outside of proto we can use libc::pid_t which is int32, so i think that's the "correct" type.

auraed/src/cells/cell_service/cell_service.rs
dmah42: ack

auraed/src/cells/cell_service/cell_service.rs
dmah42: i couldn't find a way to have a "non-blocking error".. maybe i should just catch here and warn?  that would show up in the auraed logs.

auraed/src/observe/observe_service.rs
dmah42: perhaps not.. same as above.. maybe catch and warn instead?

future-highway: I think catch and warn is a good choice

future-highway (approved):

PR #411, opened 2023-02-15
auraed ebpf refactoring

future-highway:
So far, I have been unable to get the tests to run (they are ignored even when adding the --include-ignored flag).

I'll add additional comments inline


auraed/src/ebpf/tracepoint_programs/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+\* -------------------------------------------------------------------------- */
+
+use aurae_ebpf_shared::Signal;
+pub use perf_event_listener::PerfEventListener;
+pub use tracepoint_program::TracepointProgram;
+
+mod perf_event_listener;
+mod tracepoint_program;
+
+pub struct SignalSignalGenerateTracepointProgram;
```
future-highway, line 38: I thought of naming this `PosixSignalTracepointProgram`, but I assume there was a reason for the naming we had?

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
```diff
@@ -0,0 +1,130 @@
...
+        // Load the eBPF TracePoint program
+        let program: &mut TracePoint = bpf
+            .program_mut(Self::PROGRAM_NAME)
+            .ok_or_else(|| anyhow::anyhow!("failed to get eBPF program"))?
+            .try_into()?;
+
+        // Load the program
+        match program.load() {
+            Ok(_) => Ok(()),
+            Err(ProgramError::AlreadyLoaded) => Ok(()),
```
future-highway, line 39: Seemed like it would be ok to ignore this error

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
```diff
@@ -0,0 +1,130 @@
...
+
+pub trait TracepointProgram<T: Clone + Send + 'static> {
+    const PROGRAM_NAME: &'static str;
+    const CATEGORY: &'static str;
+    const EVENT: &'static str;
+    const PERF_BUFFER: &'static str;
+
+    fn load_and_attach(
+        bpf: &mut Bpf,
+    ) -> Result<PerfEventListener<T>, anyhow::Error> {
```
future-highway, line 27: I'll refactor out the use of anyhow in this PR or a future one.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
```diff
@@ -0,0 +1,130 @@
...
+        let per_cpu_buffer_capacity = (PER_CPU_BUFFER_SIZE_IN_PAGES
+            * page_size as usize)
+            / event_struct_size;
+
+        // Set the capacity of the channel to the combined capacity of all the
+        // per-CPU buffers
+        let channel_capacity = per_cpu_buffer_capacity * num_cpus;
+
+        // Attach to kernel trace event
+        let _ = program.attach(Self::CATEGORY, Self::EVENT)?;
```
future-highway, line 63: I didn't ignore the `AlreadyAttached` error that this can throw. Should I?

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
```diff
@@ -0,0 +1,130 @@
...
+                        error!(
+                            "buffer full, dropped {} perf events - this should never happen!",
+                            events.lost
+                        );
+                    }
+
+                    for buf in buffers.iter_mut().take(events.read) {
+                        let ptr = buf.as_ptr() as *const T;
+                        let signal = unsafe { ptr.read_unaligned() };
+                        if per_cpu_tx.send(signal).is_err() {
```
future-highway, line 118: If there are no listeners, I think it's safe to ignore the error without having the `error!` trace.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
```diff
@@ -0,0 +1,130 @@
...
+                            // send only errors if there are no listeners,
+                            // it can be safely ignored; future sends may succeed
+                        }
+                        buf.clear();
```
future-highway, line 122: I don't think there was an issue with the loop re-reading signals, but I thought it would be a good idea to clear the signals after they are processed

auraed/src/lib.rs
```diff
@@ -199,24 +205,38 @@ pub async fn run(
...
+            let mut bpf = match AuraeBpf::load() {
+                Ok(bpf) => Some(bpf),
+                Err(e) => {
+                    warn!("eBPF: Cannot load probes: {e}");
+                    None
+                }
+            };
+
+            let posix_signals_listener = if let Some(bpf) = &mut bpf {
+                let posix_signals_listener = match bpf.load_and_attach_tracepoint_program::<SignalSignalGenerateTracepointProgram, _>() {
```
future-highway, line 226: If we map each program to a gRPC service, the health endpoint can be used to check if the program is loaded by checking if the service is being served.

ebpf/Makefile
```diff
@@ -42,13 +43,15 @@ release: nightly bpf-linker
...
-		rustup toolchain install nightly-x86_64-unknown-linux-gnu --component rust-src
+	rustup toolchain list | grep -qc 'nightly-$(uname_m)-unknown-linux-gnu' || \
+		rustup toolchain install nightly-$(uname_m)-unknown-linux-gnu --component rust-src
 
 .PHONY: bpf-linker
 bpf-linker: ## Install bpf-linker (needed for eBPF)
-	cargo install --list | grep -qc 'bpf-linker' || \
-		cargo install bpf-linker
+	$(cargo) install --list | grep -qc 'bpf-linker' || \
+	if [ '$(uname_m)' = 'aarch64' ]; \
```
future-highway, line 52: In the main Makefile we use `ifeq`; I couldn't get that to work, though I think that would be nicer.

auraed/src/ebpf/tracepoint_programs/mod.rs
JeroenSoeters: The event we are tracing is `signal:signal_generate` where `signal` is the subsystem and `signal_generate` is the actual event. This name makes this mapping to the Linux trace event obvious, but I agree it's not very easy on the eye. Once we have a bunch more of these tracepoints we could put them in a module per subsystem so we can get rid of the subsystem prefix.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
JeroenSoeters: Let's add a log line, so we know when something fishy is going on. We shouldn't be trying to load the same program twice.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
JeroenSoeters: Same as above ☝️

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
JeroenSoeters: When would this be the case though? 🤔 

FWIW with future tracepoints we are going to have to be a bit more mindful wrt resource consumption. This listener is continuously listening for signals, which is fine for now. But if we're instrumenting all syscalls on a machine I think we should consider just running that probe in the context of the gRPC request. Or maybe share a BPF map between kernel and userspace that has some flags which "features" we want enabled which the probe then checks before running any instrumentation code.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
JeroenSoeters: As long as it's fast :) Again for POSIX signals this won't be a hot path, but for other future instrumentations, it will be.

auraed/src/lib.rs
JeroenSoeters: This won't be the case. As we speak, I am adding at least 2 but likely more tracepoints just for the proper functioning of the signals endpoint. I do think we need some notion around what is loaded/supported at some point, as various eBPF programs will have different minimum kernel requirements, so we will likely expose a larger feature-set on more recent kernels.

I imagine at some point, we want to have remote eBPF deployment and instrumentation. Sort of an API for `bpftool` that tells you which programs are loaded, which maps are there etc. Maybe let's revisit health at that point and right now stick with "if it didn't error out, it loaded successfully"?

JeroenSoeters (changes requested):
Just a few minor comments around adding log statements when we see errors that really shouldn't be happening as long as we write proper code :)

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
future-highway: As far as I see, having no listeners is the default case. The only time we have a listener is when a client subscribes via `ObserveService::get_posix_signals_stream`

JeroenSoeters: Ah ok, I thought `send()` would only error if the channel was closed, and didn't care if there were receivers or not. Wouldn't we be seeing tons of error messages otherwise when running the daemon?

Related: maybe `PerfEventListener` is confusing, as it's really a wrapper around a `Sender<T>`.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
future-highway: I'll take it out an leave a comment

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
future-highway: I think LogChannel is the only other place we use broadcast channels and we also ignore the error for that, so we don't have any tracing of these errors.

RE: `PerfEventListener`...I assumed there would be a struct that could be subscribed, but wouldn't expose send in tokio, but didn't find one, so I actually made another struct (which I called `Subscribable`, because names are hard) before realizing you did the same in `PerfEventListener`, so I think it's the right decision to have it.

future-highway:
My thought was to refactor some more in this PR, but considering the above comments re 2 more tracepoints, I think it's better to see what happens with those first, so I'm marking this as ready for review.

@JeroenSoeters Can you run the tests as part of reviewing this?

JeroenSoeters:
@future-highway tests are passing here on my machine

JeroenSoeters:
Just a few minor naming nit-picks. IMO we should merge this ASAP, as I'm about to add a few more eBPF probes.

JeroenSoeters (approved):
Some minor nit-picks and we have some conflicts now but other than that g2g imo.

auraed/src/ebpf/tracepoint_programs/tracepoint_program.rs
future-highway: Hmm, maybe 'PerfEventBroadcast`...I think I'll update to that if there are no objections.

ebpf/Makefile
future-highway: x86_64 is the only officially supported arch as of right now so we don't need to worry about CI. I just wanted to keep the M1 as a viable dev environment, see as that's what I have :)

JeroenSoeters (approved):

PR #412, opened 2023-02-16
Add e2e-test-all target to Makefile

JeroenSoeters:
Small quality of life improvement

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,10 +240,17 @@ ebpf:
 	cd ebpf && make release install
 
 .PHONY: e2e-test
-e2e-test: # Run the end-to-end tests for auraed
+e2e-test: ## Run the end-to-end tests for auraed
 	sudo -E auraed &
 	sleep 5
 	$(cargo) test --package e2e-tests --lib -- --exact --nocapture --test-threads=1
+	sudo pkill -9 auraed
+
+.PHONY: e2e-test-all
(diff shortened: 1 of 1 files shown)


krisnova (approved):

Makefile
```diff
@@ -240,12 +240,19 @@ ebpf:
...
+e2e-test: ## Run the end-to-end tests for auraed
 	sudo -E auraed &
 	sleep 5
 	$(cargo) test --package e2e-tests --lib -- --exact --nocapture --test-threads=1
 	sudo pkill -9 auraed
 
+.PHONY: e2e-test-all
+e2e-test-all: ## Run all end-to-end tests for auraed (includes ignored tests)
+	sudo -E auraed &
+	sleep 5
```
dmah42, line 252: this feels like it's going to be a maintenance nightmare if tests start taking longer. is there no way to run auraed as part of the e2e test and keep it hermetic?

dmah42 (changes requested):

Makefile
JeroenSoeters: So the 5s here is just one time just so the tests don't start running until `auraed` is fully started but I agree this whole setup is not ideal (and if anything we shouldn't be sleeping in tests but polling). We should investigate leveraging @future-highway's work on integration tests for e2e tests so we don't have to start and stop `auraed` from the Makefile.

dmah42: +1

PR #415, opened 2023-02-17
Make ebpf-shared::Signal pid field a `u32`

izissise:
It is explicitly a u32 in `/ebpf` and u32 makes more sense for PIDs


izissise:
Arf, maybe it should be a uint32 in the protobuf?

PR #417, opened 2023-02-17
Make pid i32

izissise:
This fix cargo compilation in /ebpf folder
```
cargo +nightly build --package ebpf-probes
--target=bpfel-unknown-none -Z build-std=core --release
```

### DIFF
--- a/ebpf/src/probe-tracepoint-signal-signal-generate.rs
+++ b/ebpf/src/probe-tracepoint-signal-signal-generate.rs
@@ -65,7 +65,7 @@ fn try_signals(ctx: TracePointContext) -> Result<u32, u32> {
         }
     };
 
-    let pid: u32 = unsafe {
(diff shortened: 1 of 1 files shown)


izissise:
@dmah42 This works better

ebpf/src/probe-tracepoint-signal-signal-generate.rs
```diff
@@ -65,7 +65,7 @@ fn try_signals(ctx: TracePointContext) -> Result<u32, u32> {
         }
     };
 
-    let pid: u32 = unsafe {
+    let pid: i32 = unsafe {
```
future-highway, line 68: @JeroenSoeters I know its the same size, but is this safe to change?

future-highway (approved):
Ah, this is a requirement to build because of the ebpf-shared change.

PR #420, opened 2023-02-18
Socket connection

dmah42:
this PR started out with the bit that changes the Client to connect to a SocketAddr instead of a Uri.

but we had an adventure fixing some issues that failed build on syncing to head, and added logging when failing to parse the config file.


future-highway (approved):
LGTM

future-highway (approved):

PR #421, opened 2023-02-19
MUSL target, dependencies compiler checker

GANitak:
Here i don't know if that the way to pull in the target checks, i just looked at how future highway did it with the buf one and added the Filesystem Layout Dependencies for musl 
* But line 179 is i *suppose* is where this checker should be added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,8 +176,7 @@ PROGS = aer auraed auraescript
 define AURAE_template =
 .PHONY: $(1)
 $(1): musl $(GEN_RS) $(GEN_TS) $(1)-lint $(1)-debug ## Lint and install $(1) (for use during development)
-@musl --version >/dev/null & /dev/tty & /tmp & /bin/sh & /proc $(syslibdir)/ld-musl-$(ARCH).so.1 2>&1 || (echo "Warning: musl is not installed! Please install the target 'musl' dependencies compiler: https://doc.rust-lang.org/rustc/platform-support.html"; exit 1)
-
+	
 .PHONY: $(1)-lint
 $(1)-lint: musl $(GEN_RS) $(GEN_TS)
 	$$(cargo) clippy $(2) -p $(1) --all-features -- -D clippy::all -D warnings
@@ -216,6 +215,8 @@ $(1)-release: musl $(GEN_RS) $(GEN_TS) $(1)-lint $(1)-test ## Lint, test, and in
 endef
 
 MUSL_TARGET=--target $(uname_m)-unknown-linux-musl
+$(musl) | grep -qc 'x86_64-unknown-linux-musl' || \
+		target musl install $(uname_m)-x86_64-unknown-linux-musl 2>&1 || (echo "Warning: musl is not installed! Please install the target 'musl' dependencies compiler: https://doc.rust-lang.org/rustc/platform-support.html"; exit 1)
 
 $(foreach p,$(PROGS),$(eval $(call AURAE_template,$(p),$(if $(findstring auraed,$(p)),$(MUSL_TARGET),))))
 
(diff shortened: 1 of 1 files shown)


future-highway (changes requested):
The buf check is different, you should ignore that for this.

The goal is to put a check on the musl target in the main Makefile. We have another Makefile in the project (under the ebpf directory) that does a check using grep.

Feel free to ping in discord if you want more specifics.

Makefile
```diff
@@ -176,7 +177,7 @@ PROGS = aer auraed auraescript
 define AURAE_template =
 .PHONY: $(1)
 $(1): musl $(GEN_RS) $(GEN_TS) $(1)-lint $(1)-debug ## Lint and install $(1) (for use during development)
-
+
```
future-highway, line 180:
```suggestion

```

Makefile
```diff
@@ -116,7 +116,8 @@ config: certs ## Set up default config
 
 .PHONY: musl
 musl: ## Add target for musl
-	rustup target add $(uname_m)-unknown-linux-musl
+	rustup target list | grep -qc $(uname_m)-unknown-linux-musl || \
```
future-highway, line 119: This line is really close. But:

1. grep takes a string for the pattern input (run `grep --help`)
2. when you run `rustup target list` you'll notice it lists all targets (not just those that are installed). The ones that are installed are specifically indicated

Makefile
```diff
@@ -116,7 +116,8 @@ config: certs ## Set up default config
...
+	rustup target install $(uname_m)-unknow-linux-muslknown-linux-musl
```
future-highway, line 120: You can run `rustup target --help` to see the available subcommands. "install" is not an option.

future-highway (changes requested):
Much closer

Makefile
GANitak: so it's rustup target add

Makefile
```diff
@@ -116,7 +116,8 @@ config: certs ## Set up default config
 
 .PHONY: musl
 musl: ## Add target for musl
-	rustup target add $(uname_m)-unknown-linux-musl
+	rustup target list | grep -qc '$(uname_m)-unknown-linux-musl' || \
+	rustup target add $(uname_m)-unknow-linux-muslknown-linux-musl
```
future-highway, line 120: You're almost done. Just looks like the name of the target here got changed accidentally.

future-highway (changes requested):

Makefile
```diff
@@ -116,7 +116,8 @@ config: certs ## Set up default config
 
 .PHONY: musl
 musl: ## Add target for musl
-	rustup target add $(uname_m)-unknown-linux-musl
+	rustup target list | grep -qc '$(uname_m)-unknown-linux-musl' || \
```
future-highway, line 119: The fix to making it a string was correct, but the string itself is a little off. Installed targets (what we want grep to look for) end in " (installed)"

future-highway (changes requested):

Makefile
```diff
@@ -116,7 +116,8 @@ config: certs ## Set up default config
 
 .PHONY: musl
 musl: ## Add target for musl
-	rustup target add $(uname_m)-unknown-linux-musl
+	rustup target list | grep -qc '$(uname_m)-unknown-linux-musl (installed)' || \
+	rustup target add $(uname_m)-unknow-linux-musl
```
future-highway, line 120:
```suggestion
	rustup target add $(uname_m)-unknown-linux-musl
```

future-highway (changes requested):

future-highway (approved):
Got it!

PR #423, opened 2023-02-20
Disable default seccomp profile in CI

future-highway:
This is an attempt to get more tests running in CI

### DIFF
--- a/.github/workflows/199-release-make-ci-upload-release-artifacts.yml
+++ b/.github/workflows/199-release-make-ci-upload-release-artifacts.yml
@@ -45,6 +45,7 @@ jobs:
           credentials:
             username:  ${{ github.actor }}
             password:  ${{ secrets.GITHUB_TOKEN }}
+          options: --security-opt seccomp=unconfined
         permissions:
           contents: write
           packages: write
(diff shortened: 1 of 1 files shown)


.github/workflows/199-release-make-ci-upload-release-artifacts.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           credentials:
             username:  ${{ github.actor }}
             password:  ${{ secrets.GITHUB_TOKEN }}
+          options: --security-opt seccomp=unconfined
```
dmah42, line 48: from the discord: maybe "--privileged" is the right flag as it grants CAP_SYS_ADMIN: https://docs.docker.com/engine/reference/commandline/run/#:~:text=The%20%2D%2Dprivileged%20flag%20gives,like%20running%20Docker%20within%20Docker.

future-highway: Neither one worked. It may have to do with the how the `skip_if_seccomp` macro is detecting Seccomp mode. Going to close the PR.

future-highway:
Attempts failed. The attempted flags must not be disabling Seccomp mode, but maybe using Secomp_filters (with no acutal filters)? Closing this PR so it doesn't linger. We should revisit at some point.

PR #425, opened 2023-02-21
Add tests for cgroups validations

dmah42:


auraed/src/cells/cell_service/cells/cgroups/cpuset/cpus.rs
```diff
@@ -76,12 +67,9 @@ mod tests {
     #[test]
     fn test_validation_success(input: &str) {
         let input: Cpus = Cpus::new(input.into());
```
future-highway, line 69: I know this is from before, but it seems weird that we use `Cpus::new` and then call `.into_inner()` on ln 71. Why don't we just call `input.to_string()` on 71?

We do this in other tests as well.

auraed/src/cells/cell_service/validation.rs
```diff
@@ -294,3 +294,224 @@ impl From<ValidatedExecutable> for super::executables::ExecutableSpec {
...
+            "field",
+            Some("parent"),
+        );
+        assert!(validated.is_err());
+    }
+
+    #[test]
+    fn test_cell_type_cpu_weight_too_large() {
+        let validated = CellValidator::validate_cpu(
+            Some(CpuController { weight: Some(20000), max: None }),
```
future-highway, line 338: Is it better to test the first invalid value (i.e., 10001)?

auraed/src/cells/cell_service/executables/executable_name.rs
```diff
@@ -6,6 +6,13 @@ use validation::{ValidatedField, ValidationError};
 #[derive(Debug, Clone, Ord, PartialOrd, Eq, PartialEq, Hash)]
 pub struct ExecutableName(String);
 
+#[cfg(test)]
+impl ExecutableName {
+    pub(crate) fn new(name: String) -> Self {
```
future-highway, line 11: I think this should have the `#[cfg(test)]` attribute.

future-highway (commented):
Comment re `#[cfg(test)]` is the only one that stopped my approving.

krisnova:
Can we please address @future-highway's requests and then we can merge this in?

auraed/src/cells/cell_service/executables/executable_name.rs
dmah42: it doesn't inherit from the impl?

auraed/src/cells/cell_service/validation.rs
dmah42: a QA tester walks into a bar, orders 1 beer, 10001 beers, 20000 beers, NaN beers...

auraed/src/cells/cell_service/executables/executable_name.rs
future-highway: I think you are right. It should inherit; I missed that it was above. I do prefer it on the fn though (but wouldn't have not approved had I seen it).

future-highway (approved):

PR #426, opened 2023-02-22
Use /proc/self/exe to execute nested auraed

izissise:
Use /proc/self/exe symlink to execute another auraed process. Previously, if auraed binary isn't in $PATH commands lookup would not find the auraed binary.


future-highway (approved):
Great! Thank you

PR #427, opened 2023-02-24
Auraescript create one client and give it to services

izissise:
TODO:
- [x] One client for all methods?
- [x] Use `AuraeConfig` struct
- [x] Fallback to default values if no args to constructor
- [x] Can give a toml config file?

Maybe we only need one client instance and give it to each service instance?
```
let client = AuraeClient({
   ...
});
let cells = new runtime.CellServiceClient(client);

...
```


examples/cells_output.ts
```diff
@@ -31,7 +31,7 @@
 import * as helpers from "../auraescript/gen/helpers.ts";
 import * as runtime from "../auraescript/gen/cells.ts";
 
-let cells = new runtime.CellServiceClient();
+let cells = new runtime.CellServiceClient("./pki/ca.crt", "./pki/_signed.client.nova.crt", "./pki/client.nova.key", "127.0.0.1:8080");
```
dmah42, line 34: super nit: if the socket could be a socketaddress and we could take advantage of typescript's type safety that would be neat.  i don't know if that's even possible (there's an "ip-address" node module but i don't know if that's importable through typescript+deno or how it translates to a type in rust).

dmah42, line 34: given all the clients would want to use the same certs+socket (right?) i think this should be a higher level API ... otherwise we'll be passing the same thing to every client.

izissise: Maybe it could even be a [Deno.Conn](https://deno.land/api@v1.31.0?s=Deno.Conn), I wonders if we can give tonic client an already open socket

izissise:
V2 with one client for all services, it wasn't easy to figure out how deno's resource_table work.
In the end, deno macro does a magic thing were OpState is already filled when calling from typescript

examples/cells_output.ts
dmah42: i _think_ i'm more comfortable passing data from TS to auraescript for it to open the connection, similarly to how `ae` would do it.  finding the balance between "rich" script and client is tricky but opening the connection crosses the line for me (but this is only my opinion).

auraescript/helpers.js
```diff
@@ -14,3 +14,10 @@ function toString(value: any): string {
         return value?.toString();
     }
 }
+
+function auraescript_client(
```
dmah42, line 18: nit: just "client"? or "createClient"?

auraescript/src/bin/main.rs
```diff
@@ -65,6 +65,9 @@ fn main() -> anyhow::Result<()> {
 
     let _ = js_runtime.execute_script("", "Deno.core.initializeAsyncOps();")?;
 
+    // include helper.ts
+    let _ = js_runtime.execute_script("", include_str!("../../helpers.js"))?;
```
dmah42, line 69: if this is now some sort of core runtime perhaps it should be called "core.js" or "aurae.js" or something?

examples/cells_output.ts
```diff
@@ -28,10 +28,12 @@
...
  *                                                                            *
 \* -------------------------------------------------------------------------- */
-import * as helpers from "../auraescript/gen/helpers.ts";
 import * as runtime from "../auraescript/gen/cells.ts";
 
-let cells = new runtime.CellServiceClient();
+
+let client = await auraescript_client("./pki/ca.crt", "./pki/_signed.client.nova.crt", "./pki/client.nova.key", "127.0.0.1:8080");
+// helpers.print(client)
+let cells = new runtime.CellServiceClient(client);
```
dmah42, line 36: this is great.

future-highway:
I'm not a fan of the changes in the "Make helpers included in global scope" commit.

While it is nice to not have to write `import * as helpers from "../auraescript/gen/helpers.ts";` at the top of every script, we've lost the help of the IDE and typescript's type system.

I do think dma's suggestion of renaming helpers is a good idea though as the file is now providing more than just helpers/utility functions.

examples/cells_output.ts
izissise: Looking a Deno 's op interface I don't think it is possible to transfer complex object from js to rust

auraescript/src/bin/main.rs
izissise: I like aurae.js.
I'll try to keep it as typescript and compile it to js in build.rs

izissise: Renamed to aurae.js

auraescript/helpers.js
izissise: Done

izissise:
It should be good for a new review

examples/cells_quota.ts
```diff
@@ -28,10 +28,11 @@
  *   limitations under the License.                                           *
  *                                                                            *
 \* -------------------------------------------------------------------------- */
-import * as helpers from "../auraescript/gen/helpers.ts";
+import * as aurae from "../auraescript/gen/aurae.ts";
 import * as runtime from "../auraescript/gen/cells.ts";
```
dmah42, line 32: while you're here, can this be `import * as cells`?

auraescript/src/builtin/auraescript_client.rs
```diff
@@ -0,0 +1,82 @@
+use anyhow::Result;
+use client::{Client, AuraeConfig};
+use deno_core::{OpState, Resource, ResourceId};
+use std::{cell::RefCell, rc::Rc};
+
+// `AuraeConfig` `try_default`
+#[deno_core::op]
+pub(crate) async fn as__aurae__config__try_default(
```
future-highway, line 8: I'm trying to understand the naming convention here. I see you used  "as" instead of "ae" (makes sense since these ops are auraescript specific), but I'm not sure what the logic for the middle is. I would think this would be `as__builtin__client__try_default` and that the others would follow the same pattern.

izissise: I choose aurae_config because try_default is a `AuraeConfig` method

auraescript/src/bin/main.rs
```diff
@@ -65,6 +65,9 @@ fn main() -> anyhow::Result<()> {
 
     let _ = js_runtime.execute_script("", "Deno.core.initializeAsyncOps();")?;
 
+    // include aurae.ts
+    // let _ = js_runtime.execute_script("", include_str!(concat!(env!("CARGO_MANIFEST_DIR"), "/gen/aurae.ts")))?;
```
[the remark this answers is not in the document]
izissise: Yep shouldn't have been left here

auraescript/aurae.ts
```diff
@@ -0,0 +1,35 @@
...
+        value !== null
+    ) {
+        return JSON.stringify(value, null, 2);
+    } else {
+        return value?.toString();
+    }
+}
+
+export function createClient(
+    path_or_ca_crt?: string, client_crt?: string, client_key?: string, socket?: string,
```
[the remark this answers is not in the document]
izissise: Yes that's much better, I wonder if it could be generated from `AuraeConfig` struct

examples/cells_quota.ts
izissise: I agree this would look better, but as @future-highway said (https://github.com/aurae-runtime/aurae/pull/427#issuecomment-1445377988) changing the import will break IDE support
I didn't find a way to hint typescript LSP for globally existing functions or import paths, I haven't looked too hard though

future-highway: I think dma was suggesting `import * as cells from "../auraescript/gen/cells.ts"`.

What is called cells.ts used to be runtime.ts and we missed a couple of renames.

auraescript/src/builtin/auraescript_client.rs
future-highway: Ah, that makes sense. So this is nit: there is an extra "_" between `aurae` and `config`, IMO ( `as__aurae_config__try_default`)

examples/cells_quota.ts
izissise: Oh! I get it now

auraescript/macros/src/ops.rs
```diff
@@ -209,7 +220,14 @@ fn typescript_service_generator(
 ) -> String {
     let service_name = service.name();
     let mut ts_funcs: String = format!(
-        "export class {service_name}Client implements {service_name} {{"
+        r#"
+export class {service_name}Client implements {service_name} {{
+    client: u32
+
+    constructor(client: u32) {{
```
future-highway, line 227: Should we use constructor overloading like [here](https://stackoverflow.com/a/12702786)?

I'm trying to think of a way of not forcing every script to have to create the client. The op_ident above can be changed to take `client: Option<::deno_core::ResourceId>`, in which case it can fall back to using `::client::Client::default().await?;`.

Is this a good idea?

auraescript/aurae.ts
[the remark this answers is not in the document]
izissise: I have something that works but that's not really nice typescript :/

auraescript/macros/src/ops.rs
izissise: I tried what you described in the last commit, and it seem to work great  :100:

auraescript/aurae.ts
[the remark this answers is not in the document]
future-highway: It looks like your changes match exactly with https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions. I think the change is good, but I don't do much TS/JS, so maybe I'm missing something.

auraescript/aurae.ts
```diff
@@ -0,0 +1,65 @@
...
+        // resolve kind
+        if ("path" in opts) {
+            opts.kind = "path";
+        } else if ("ca_crt" in opts) {
+            opts.kind = "opts";
+        }
+    }
+    let config;
+    switch(opts.kind) {
+        case "default":
```
future-highway, line 48: Linter is asking that the case/default blocks be wrapped in `{...}`

auraescript/aurae.ts
```diff
@@ -0,0 +1,65 @@
...
+    client_crt: string;
+    client_key: string;
+    socket: string;
+};
+type CreateClient =
+    | CreateClientDefault
+    | CreateClientPath
+    | CreateClientOpts;
+
+export function createClient(opts: CreateClient = { kind: "default" }): AuraeClient {
```
future-highway, line 37: The return type needs to be wrapped in a `Promise`, so either `Promise<AuraeClient>` or `Promise<number>`, depending on how you handle my feedback in the other comment.

auraescript/macros/src/ops.rs
```diff
@@ -78,17 +78,30 @@ pub(crate) fn ops_generator(input: TokenStream) -> TokenStream {
...
+                            req: ::proto::#module::#input_type,
                         ) -> std::result::Result<
                             ::proto::#module::#output_type,
                             ::anyhow::Error
                         > {
-                            let client = ::client::Client::default().await?;
+                            let client = match client_rid {
+                                None => ::deno_core::RcRef::new(::client::Client::default().await?),
+                                Some(client_rid) => {
+                                    let asclient = {
```
future-highway, line 95: Nit: `as_client`

auraescript/macros/src/ops.rs
```diff
@@ -78,17 +78,30 @@ pub(crate) fn ops_generator(input: TokenStream) -> TokenStream {
                     let output_type = Ident::new(output_type, file_path_span);
                     let name = Ident::new(&m.name().to_snake_case(), file_path_span);
 
+                    // Magic OpState from deno (https://github.com/denoland/deno/blob/b6ac54815c1bcfa44a45b3f2c1c982829482477f/ops/lib.rs#L295)
                     quote! {
                         #[::deno_core::op]
                         pub(crate) async fn #op_ident(
-                            req: ::proto::#module::#input_type
+                            opstate: Rc<RefCell<OpState>>, // Auto filled by deno macro, call from typescript ignoring this parameter
```
future-highway, line 85: Nit: `op_state`

Like the other comment, but this one is less bothersome since my linter doesn't care about it, lol.

izissise:
I'll install typescript LSP that should ease the process

auraescript/aurae.ts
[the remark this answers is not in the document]
izissise: I'll settle for number for now

auraescript/aurae.ts
izissise: I had to keep the `break;` or the linter would complain about the last line not being reachable

examples/cells.ts
```diff
@@ -28,61 +28,62 @@
...
-import * as helpers from "../auraescript/gen/helpers.ts";
-import * as runtime from "../auraescript/gen/cells.ts";
+import * as aurae from "../auraescript/gen/aurae.ts";
+import * as cells from "../auraescript/gen/cells.ts";
 
-let cells = new runtime.CellServiceClient();
-const cellName = "ae-sleeper-cell";
+let client = await aurae.createClient();
+let cellService = new cells.CellServiceClient(client);
+const cellName = "ae-sleeper-cell2";
```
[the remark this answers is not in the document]
izissise: oops nice catch

future-highway (dismissed):
LGTM. Great job!

izissise:
Rebased

future-highway (approved):

PR #429, opened 2023-02-25
Add UTM virtualization option to M1 documentation

StephenWeber:
My own notes on setting up a VM using UTM on my own Apple Silicon computer. I expect this to take some rounds of editing / proofing, so please do ask questions and have me expand on anything. I'd really like to get the tone / audience-expectations right.

This is what worked for me, on my machine, to get through the second Virtual Machine section. I was indeed able to build aurae. 

Potential areas of improvement:
- I'm not using CLion so I don't know if I've introduced issues there?
- I never got shared-clipboard working, even when following UTM's instructions and installing their guest support
- While the `chown -R $USER:$USER /share` part works, it doesn't cover new files and I've had to re-run that especially after pulling main. There's probably a better `fstab` configuration but I'm not great at that.
- I wasted a lot of time on the shared folder trying different things and could have omitted a necessary step.
- Since I followed the Parallels instructions to set up my UTM VM, I may have omitted some of those necessary steps from  my own notes. One example is the recommendations on resources to assign to the VM (cores, RAM, disk).

Screenshot 2023-02-24 at 5 32 45 PM


StephenWeber:
@cla-bot check

krisnova:
My advice would be to do the following:

 - Define what the **UTM** acronym is in our docs (Its not clear to me as a non-apple user). We need folks to understand if something is relevant to them or not.
 - Include clear language about your work and your intentions in each PR. Do you intend to follow up in the future on your docs so its safe to merge now? Or will this be your only contribution so we need to make sure we get more in this work? 

Also please consider using project native workflows whenever possible. I don't want to live in a world where everyone is having to guess/check their bespoke workflow for the project. If you need to mount something in a VM, chances are you aren't alone so the discussion we should be having is "What is the right way for the project to mount <thing> in a VM" at large.

krisnova:
Closed for https://github.com/aurae-runtime/aurae/pull/463

PR #432, opened 2023-02-26
Macro for testing async code

JeroenSoeters:
This PR gets rid of the `sleep`'s we had in the `ProcCache` by introducing a `assert_eventually_eq` macro.


krisnova (approved):
I think in the future we might want to make the `left` and `right` durations configurable but for now this is great!

PR #438, opened 2023-02-27
In makefile, don't use sudo if already root

future-highway:


Makefile
```diff
@@ -290,7 +306,11 @@ docs-other:
 
 .PHONY: docs-serve
 docs-serve: docs ## Run the aurae.io static website locally
+ifeq ($(uid), 0)
+	./hack/serve.sh
+else
 	sudo -E ./hack/serve.sh
```
dmah42, line 312: this shouldn't need to be root i think?

Makefile
```diff
@@ -198,15 +198,27 @@ $(1)-test: musl $(GEN_RS) $(GEN_TS)
 
 .PHONY: $(1)-test-all
 $(1)-test-all: musl $(GEN_RS) $(GEN_TS)
+ifeq ($(uid), 0)
+	$(cargo) test $(2) -p $(1) -- --include-ignored
```
dmah42, line 202: should we define a `root_cargo = sudo -E $(cargo)` and use that everywhere that we need root cargo?

Makefile
future-highway: I'm having trouble testing this as it needs protoc-gen-doc and I can't seem to get it correctly installed.

dmah42 (approved):

PR #440, opened 2023-03-01
Cargo deny check licenses

future-highway:
With these changes we will be passing the cargo-deny checks for licenses.

I left advisories off of CI for now. We fail the check.

Our ebpf-shared crate did not have a license. I set it to Apache 2.0, like auared. Should it be MIT since the ebpf crate is Dual MIT/GPL?

### DIFF
--- a/.github/workflows/206-deny-check-aurae-builder-make-check-deny.yml
+++ b/.github/workflows/206-deny-check-aurae-builder-make-check-deny.yml
@@ -27,65 +27,51 @@
 #   limitations under the License.                                             #
 #                                                                              #
 # ---------------------------------------------------------------------------- #
-#
-# The primary Aurae test container, used to cache and support the GitHub
-# builds.
-#
-# This container image is NOT intended to run Aurae in production.
-# For internal builds of auraed use Dockerfile.build
-#
-# syntax = docker/dockerfile:1.4
-FROM rust:1-slim-bullseye as main
-LABEL org.opencontainers.image.source https://github.com/aurae-runtime/aurae
 
-## Define ARGs
-ARG CACHE_VERSION=v0
-ARG BUF_VERSION=1.11.0
-ARG VALE_VERSION=2.21.3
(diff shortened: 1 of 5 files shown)


.github/workflows/206-deny-check-aurae-builder-make-check-deny.yml
```diff
@@ -0,0 +1,78 @@
...
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    strategy:
+      matrix:
+        checks:
+#          - advisories
+          - bans licenses sources
+    # Prevent sudden announcement of a new advisory from failing ci:
+    continue-on-error: ${{ matrix.checks == 'advisories' }}
```
dmah42, line 63: with continue-on-error in place can we enable advisories?  it might be good to have them logged so we can think about starting to fix them.

Makefile
```diff
@@ -392,7 +392,11 @@ headers-write: ## Fix any problematic files blindly.
...
-	cargo +nightly udeps --package client
-	cargo +nightly udeps --package aer
+	$(cargo) +nightly udeps --target $(uname_m)-unknown-linux-musl --package auraed
+	$(cargo) +nightly udeps --package auraescript
+	$(cargo) +nightly udeps --package client
+	$(cargo) +nightly udeps --package aer
+
+.PHONY: check-deny
+check-deny: musl $(GEN_TS) $(GEN_RS) ## Run cargo-deny
+	$(cargo) deny check
```
dmah42, line 402: does this not run "advisories"?

.github/workflows/206-deny-check-aurae-builder-make-check-deny.yml
future-highway: I was thinking we get it passing before doing that, but I like your point more.

Makefile
future-highway: It does, but it is not what runs in CI. I should update the workflow name.

PR #441, opened 2023-03-05
Don't use /proc/self/exe in tests

future-highway:
Test binaries do not include the code in main.rs, and therefore, do not have command arg parsing. So, when we use `/proc/self/exe` to launch a nested auraed in tests, the command args are not parsed and the nested auraed runs, but doesn't do anything.

### DIFF
--- a/auraed/src/lib.rs
+++ b/auraed/src/lib.rs
@@ -101,6 +101,9 @@ mod observe;
 mod spawn;
 mod vms;
 
+#[cfg(not(test))]
+const PROC_SELF_EXE: &str = "/proc/self/exe";
+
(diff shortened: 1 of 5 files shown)


auraed/src/lib.rs
```diff
@@ -111,6 +111,8 @@ static AURAED_RUNTIME: OnceCell<AuraedRuntime> = OnceCell::new();
 /// filesystem at runtime in order to authenticate.
 #[derive(Debug)]
 pub struct AuraedRuntime {
+    /// Path to the auraed binary. Set as `None` to use the symbolic link from /proc/self/exe.
```
dmah42, line 114: instead of special behaviour for None, can we default to the symbolic link?

dmah42 (approved):

dmah42 (approved):

PR #442, opened 2023-03-07
Nested auraed inherit config through command args

izissise:
In order for relative path in options to work, we also remove current_dir("/") for the forked client

Make ClientConfig output warnings on stderr

### DIFF
--- a/auraed/src/cells/cell_service/cells/nested_auraed/nested_auraed.rs
+++ b/auraed/src/cells/cell_service/cells/nested_auraed/nested_auraed.rs
@@ -66,27 +66,39 @@ impl NestedAuraed {
         // TODO: handle expect
         let mut client_config =
             AuraeConfig::try_default().expect("file based config");
-        client_config.system.socket = format!(
+        let client_socket = format!(
             "{}/aurae-{}.sock",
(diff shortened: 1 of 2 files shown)


auraed/src/cells/cell_service/cells/nested_auraed/nested_auraed.rs
```diff
@@ -66,27 +66,39 @@ impl NestedAuraed {
...
         let mut client_config =
             AuraeConfig::try_default().expect("file based config");
-        client_config.system.socket = format!(
+        let client_socket = format!(
             "{}/aurae-{}.sock",
             auraed_runtime.runtime_dir.to_string_lossy(),
             uuid::Uuid::new_v4(),
         );
 
+        client_config.system.socket = client_socket.clone();
```
dmah42, line 75: why is this clone necessary when we only take a reference below?

izissise: `client_socket` is moved into `client_config.system.socket`, I could use a ref on `client_config.system.socket` below but I find it less readable

krisnova (approved):

PR #443, opened 2023-03-07
Remove TLS between daemon and cells

izissise:
Because the daemon should not have access to the client key,
it should not be able to reuse the client auth to communicate with
the nested_auraed. Here we disable tls for proto message between
the daemon and the cell.

This not really intended to be merged as is, but mostly to trigger the discussion
on what are the security requirements for the communication between daemon and cells.  

This pr is on top of https://github.com/aurae-runtime/aurae/pull/442

### DIFF
--- a/auraed/src/cells/cell_service/cell_service.rs
+++ b/auraed/src/cells/cell_service/cell_service.rs
@@ -64,8 +64,8 @@ macro_rules! do_in_cell {
     ($self:ident, $cell_name:ident, $function:ident, $request:ident) => {{
         let mut cells = $self.cells.lock().await;
 
-        let client_config = cells
-            .get(&$cell_name, |cell| cell.client_config())
+        let client_system_config = cells
+            .get(&$cell_name, |cell| cell.client_system_config())
             .map_err(CellsServiceError::CellsError)?;
 
         let mut retry_strategy = backoff::ExponentialBackoffBuilder::new()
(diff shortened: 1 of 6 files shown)


krisnova:
This is a fantastic conversation to be having. Thank you for bringing this up. I see two options forward (feel free to make suggestions if there is an option I am not considering that you feel is viable!). 

1. Disable TLS and have 1:1 open connections
2. Generate new TLS material and pass a fresh client certificate down to the nested cell somehow.

My first thought on 1 is that it is potentially a security hole for the project, but is probably a perfectly reasonable place to start. 

I think my decision on the topic is just that, to **start with an open and un-encrypted connection between the host daemon and the nested cell**. We can make improvements on the connection in future PRs. Maybe it makes sense to do some sort of lower level access control that only allows connections from a given socket or something. Unsure at the moment. 

If you have an opinion or suggestion on if we should generate TLS client material as described in 2. please open an issue on that topic specifically and we can introduce that change as we make a decision as a project.

izissise:
Awesome!, thanks for the response!
I'll do the following then:
- Make the current pr code more clear and we can merge once everything is right
- Open an issue for 2, I might have some idea (memfd?)

izissise:
Rebased on main

krisnova (approved):

krisnova:
Rebased in https://github.com/aurae-runtime/aurae/pull/462

PR #445, opened 2023-03-09
fix CVE RUSTSEC-2020-0071

voigt:
The version of `chrono` used by auraed depends on a vulnerable version of `time` crate. Updating `chrono` should fix this CVE.

This blocks pipeline, see [example](https://github.com/aurae-runtime/aurae/actions/runs/4359263923/jobs/7625884873#step:6:17).


dmah42:
i believe some of our dependencies might still have this dependency but hopefully this helps unblock us.

dmah42 (approved):

Issue #439, opened 2023-02-27
Implement the VmsService


JeroenSoeters:
We need a way to manage virtual machines in Aurae. The `PodSandbox` work can be layered on top of this: https://github.com/aurae-runtime/aurae/issues/433

JeroenSoeters:
With a few small changes (https://github.com/aurae-runtime/aurae/pull/447) `auraed` is now running as true PID 1 in a VM spawned with cloud-hypervisor. The VM uses a custom kernel with PVH enabled so we can do a direct kernel boot instead of a full firmware boot. For this kernel I've borrowed [cloud-hypervisor's config](https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/main/resources/linux-config-x86_64) but with IPv6 enabled. We should review those kernel options. The binaries are being copied onto the disk image now after every build.

There are a few open questions for which we already have other issues:
- what will the networking set up look like: https://github.com/aurae-runtime/aurae/issues/446
- how do we go about certificate provisioning: https://github.com/aurae-runtime/aurae/issues/328

Another big open question is how will we deploy the hypervisor(s) and which one(s) do we support.

JeroenSoeters:
Apparently we had an open issue for the kernel discussion as well :) 

https://github.com/aurae-runtime/aurae/issues/238

PR #447, opened 2023-03-09
Few fixes that allow auraed to run as true PID 1

JeroenSoeters:
This PR fixes a few things that were preventing `auraed` to run as true PID 1. This is required for the upcoming VM work.


auraed/src/init/system_runtimes/pid1_system_runtime.rs
```diff
@@ -87,9 +87,6 @@ impl SystemRuntime for Pid1SystemRuntime {
         // TODO We likely to do not need to mount these filesystems.
         // TODO Do we want to have a way to "try" these mounts and continue without erroring?
 
-        MountSpec { source: None, target: "/dev", fstype: Some("devtmpfs") }
-            .mount()?;
```
JeroenSoeters, line 91: This was throwing a "Resource busy" so I presumed this wasn't actually needed?

future-highway: Is it just a feature of cloud-hypervisor that /dev is already mounted, but we would need to mount it in a non-vm context (i.e., true-true PID 1 for lack of a better identifier)?

Maybe the try and continue approach is right?

future-highway: Looking at [cloud-hypervisor's config](https://raw.githubusercontent.com/cloud-hypervisor/cloud-hypervisor/main/resources/linux-config-x86_64) that you linked to in #439 I see the following in the file:

```
CONFIG_DEVTMPFS=y
CONFIG_DEVTMPFS_MOUNT=y
```

Would this config be causing the mount to already be mounted?

JeroenSoeters: Probably so. "Resource busy" usually means that the device is already mounted afaik

dmah42: sounds like we should try and mount just in case the config doesn't exist to automatically mount it (but allow the error if it fails... maybe log a warning)

JeroenSoeters: We should probably do that for all mounts we deem essential. I'm wondering though if we should do this now or if we can postpone this to later when we allow for kernels other than the ones we control - if we ever are planning to do this at all, as this introduces a whole set of other challenges. Until that point we know with 100% certainty what is mounted and what is not.

krisnova: Lets postpone until later once we have better control of the kernels we are running.

krisnova (approved):

Issue #398, opened 2023-02-12
Cargo.nix Why does it need to live here?


krisnova:
I understand that `Cargo.nix` is a file that "pins" the Cargo dependencies to a specific version/release/etc when compiling Aurae for Nix.

My question isn't what the file it **is** or what the file **does**... my question is why is it in our repository? 

## Few Questions

 - Is this file something that the project should support? 
 - Do we intend on supporting Nix compiling Aurae from scratch in favor of downloading a signed artifact? 
 - Do we intend to keep these operating system specific files in our core repo? 
 - Should these operating system specific configs be moved to other repos? (For example: a nix repo?)

---

Basically my concern is that I am not sure why the project needs to track 4,500+ lines of `.nix` data in our core repo? It is polluting our metrics and bloating the repository. 

![image](https://user-images.githubusercontent.com/13757818/218329718-4ece0563-833f-4578-be05-b80e11b06727.png)

hbjydev:
Just one question to answer for me :) I don't wanna get involved in "should it be here/do we wanna care" discussions because I've made exactly 1 PR lol.

> Do we intend on supporting Nix compiling Aurae from scratch in favor of downloading a signed artifact?

This just _won't work_ if you run NixOS, as Nix poisons linking for stuff it builds in a way that you can only run it with Nix. However, _NixOS_ (the operating system) relies _totally_ on this and doesn't follow the Linux FHS for the most part, so `/lib` and `/lib64`, etc. aren't really reliable. That means that anything compiled in a way other than 110% statically is unlikely to run. I've even seen some binaries that _are_ statically-linked not work on NixOS.

Someone can correct me if I'm wrong but that's been my experience.

Source for the LD stuff not working: https://github.com/NixOS/nixpkgs/issues/107375#issuecomment-749580559
This holds true still.

marcusramberg:
I was the one who opened the PR to bring it back, but having said that I'm not sure I love the cargo2nix stuff as it seems like you'd need to keep it in sync with the cargo file as aurae changes? The reason I opened it was the `flake.nix` file still in the repo was referring to it. Having said that I don't love the cargo2nix stuff as it seems like you'd have to keep this file up to date, and like Nova points out it adds a lot of auto generated code to the project. Isn't there a simpler rust builder we could use? I guess the intent when adding this was for actual development of Aurae and not distribution, however I found it easier to just use rustup inside Nixos as well.

hbjydev:
[...]

PR #448, opened 2023-03-09
nix: minimum viable dev shell for `make build` and `make test`

jhillyerd:
Background: The current cargo2nix based nix flake is broken, and cannot be used to provide a development environment to nix users, it errors out.

This PR introduces a minimal viable devShell that provides a rustup environment for building auraed and auraescript.  It  uses a shellHook to set a default rust toolchain, as the existing Makefile makes changes to the rustup environment.  We may want to switch the project to use a `rust-toolchain.toml` file in the future, but I don't know enough about aurae to properly test that.

This is a first step towards removing cargo2nix as agreed in #398.  I'll send a followup PR to remove the cargo2nix portions of the flake.

Update: Added libseccomp dependency, and confirmed all of `make build` and `make test` succeed.


jhillyerd:
@cla-bot check

ph:
I've been looking at this situation too, I am kinda new to nix. I wonder if we should really clean the file up as you mentioned and also use the `rust-toolchain.yml` directly using the overlay helpers `fromRustupToolchainFile` something like this in the flake.nix:

```nix
{
  description = "A devShell example";

  inputs = {
    nixpkgs.url      = "github:NixOS/nixpkgs/nixos-unstable";
    rust-overlay.url = "github:oxalica/rust-overlay";
    flake-utils.url  = "github:numtide/flake-utils";
  };

  outputs = { self, nixpkgs, rust-overlay, flake-utils }:
    flake-utils.lib.eachDefaultSystem (system:
      let
        overlays = [ (import rust-overlay) ];
        pkgs = import nixpkgs {
          inherit system overlays;
        };

        rustVersion = pkgs.rust-bin.fromRustupToolchainFile ./rust-toolchain.toml;
      in {
        devShell = pkgs.mkShell {
          buildInputs =
            [
              (rustVersion.override { extensions = [ "rust-src" ]; })
              pkgs.rust-analyzer
              pkgs.protobuf
              pkgs.buf
              pkgs.libseccomp
            ];
        };
      });
}
```
file: rust-toolchain.toml
```toml
[toolchain]
channel = "1.68.0"
components = [ "rustfmt", "clippy", "rust-analyzer"]
targets = [ "x86_64-unknown-linux-musl", "aarch64-unknown-linux-musl" ]
```

We would also need to patch the `Makefile` to short circuit the `rustup` calls. I used the following, Note I don't know what are the good practice in nix for theses situation. But it seems to make everything in local environment work.

```makefile
.PHONY: musl
musl: ## Add target for musl
ifeq ($(origin IN_NIX_SHELL), undefined)
	rustup target list | grep -qc '$(uname_m)-unknown-linux-musl (installed)' || \
	rustup target add $(uname_m)-unknown-linux-musl
endif
```

jhillyerd:
@ph yes, your example is very similar to what I was thinking.  Instead of hacking the Makefile to check for nix shell, it may make sense for the entire project to switch over to rust-toolchain.toml, as that's [part of rustup](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file) and not specific to nix.  I suspect the Makefile is going about things the wrong way.

All that seems a bit much for my first PR to the project, I just want to make an incremental improvement before I start deleting the broken aspects of the nix build.

krisnova (approved):

PR #450, opened 2023-03-10
Use `AuraedPath` to contain logic for locating auraed exe

future-highway:
In #447 @JeroenSoeters points out that we are reading `/proc/self/exe` too early (`/proc` may not be mounted, specifically when running as true PID 1).

The approach in this PR consolidates the logic for determining the path to auraed into a `AuraedPath` where we read `/proc/self/exe` when needed as opposed to when launching auraed.


JeroenSoeters (approved):
LGTM :+1:

PR #456, opened 2023-03-14
Add a macOS + Lima development environment guide

BradLugo:
Adds a guide for using Lima on macOS with an option to mount the auraed unix socket from the guest VM to the host machine, so a developer can interact with aurae from their host machine.


BradLugo:
Note that the PR is for the `docs-guide` branch. Happy to change this to `main` if `docs-guide` gets merged in

dmah42 (approved):

Issue #461, opened 2023-03-19
Inefficient Builds


krisnova:
I think there is opportunity for newcomers to the project to help with our current build times.

Right now our current builds can take up to 15 minutes, and I believe there is a lot of repeated and duplicated work. The more compiling and downloading and linking commands we do, the longer the builds.

How do we get build times down to just a few seconds? Maybe less than 60 seconds for the entire project? 

![image](https://user-images.githubusercontent.com/13757818/226204177-be8d4a04-2c9c-40a4-9931-1b1510ca7284.png)

krisnova:
Was able to get a 30 minute build.. ouch

dmah42:
```
$ rustup default nightly
$ export RUSTFLAGS=-Zself-profile
$ make all
```

unfortunately this build fails due to tonic generating code that fails the nightly clippy checks.

future-highway:
> naively i'd expect that if we merge these into a single crate we'd end up spending less time dealing with dependencies but having less flexibility to build things individually.

The guidance for Rust is usually to break projects up into small crates since crates that don't depend on each other can be built in parallel.

By looking at an [action](https://github.com/aurae-runtime/aurae/actions/runs/4471589583/jobs/7856671589) that missed the cache and picking a single dependency (I'm picking `serde` under the assumption that it is a common dependency; it appears 44x in our Cargo.lock), we can get some hints on how Cargo reuses its output. 

The following log lines appear for serde:
```bash
cargo build --target x86_64-unknown-linux-musl -p auraed
Downloaded serde v1.0.152
Compiling serde v1.0.152
...
cargo build  -p auraescript
Compiling serde v1.0.152
...
cargo build  -p aer
Compiling serde v1.0.152
...
cargo clippy --all-features --workspace --exclude auraed --exclude auraescript --exclude aer  -- -D clippy::all -D warnings
Checking serde v1.0.152
...
cargo clippy --target x86_64-unknown-linux-musl -p auraed --all-features -- -D clippy::all -D warnings
Checking serde v1.0.152
...
cargo clippy  -p auraescript --all-features -- -D clippy::all -D warnings
Checking serde v1.0.152
...
cargo install --target x86_64-unknown-linux-musl --path ./auraed --debug --force
Downloaded serde v1.0.158
Compiling serde v1.0.158
...
cargo install  --path ./auraescript --debug --force
Compiling serde v1.0.158
```

Notes:
- `cargo build` does a single download regardless of target
[...]

PR #465, opened 2023-03-21
Use outdated cache as backup in GHA

future-highway:
Docs for [restore-keys](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

Thinking an outdated cache is better than no cache

Hopefully helps with #461


dmah42 (approved):

future-highway:
101 only failed due to timing out when trying to cache the output, so I'm going to override and merge

PR #466, opened 2023-03-23
feat(auraescript) converts from deno_core::JsRuntime to deno_runtime::MainWorker

### LINKED ISSUE #434: Utilize `deno_runtime::MainWorker` to enable more Deno features when running AuraeScript files
Currently AuraeScript is only using the `deno_core::JsRuntime` to run AuraeScript files, which only provides core Deno functionality. In order to enable additional Deno features like Web APIs, IO, file system access, HTTP (which should all be safe to enable if AuraeScript files are only being executed on the user's machine) we can use the `deno_runtime::MainWorker` struct which bakes all of the corresponding native ops and js files needed for those to work. 

Currently switching to `deno_runtime::MainWorker` is blocked due to a lock on `uuid` version `=1.1.2` in the Deno workspace. A PR for unlocking and upgrading the `uuid` version has been approved and merged (https://github.com/denoland/deno/pull/17963), but currently waiting on a new publish for the crate.

EDIT:

Will submit a PR with the AuraeScript change once the changes to the `deno_runtime` crate are published.

seanwatters:
This change enables usage of all the APIs that Deno CLI provides (web, fs, net, etc.) in AuraeScript.

Includes upgrades to `deno_core` and `deno_ast`, and fixes associated with the version bump. 

Resolves  #434


seanwatters:
@cla-bot check

dmah42 (commented):
does this mean we can replace the logging helper typescript function with console.log out similar?

auraescript/src/lib.rs
```diff
@@ -82,14 +91,64 @@ mod discovery;
...
+    let create_web_worker_cb = Arc::new(|_| {
+        todo!("Web workers are not supported yet");
+    });
+    let web_worker_event_cb = Arc::new(|_| {
+        todo!("Web workers are not supported yet");
+    });
+
+    MainWorker::bootstrap_from_options(
+        main_module,
+        PermissionsContainer::allow_all(),
```
future-highway, line 110: I remember I said in the discord that I think all permissions are ok, but just so I get double checked on this...

Auraescript will have all [permissions](https://deno.land/manual@v1.32.0/basics/permissions) including:
- get/set env vars
- access to APIs that provide information about user's operating system
- high-resolution time measurement
- network access
- loading of dynamic libraries
- file system read access
- running subprocesses
- file system write access

@dmah42 @krisnova

dmah42: i think your first statement is correct.  as it is run by (and likely written by) a client and can be readily inspected by the client, i think this is fine.

it's like any scripting language: the right function call can expose information or write to the file system, but it's on the person running the script to understand what they're running.

seanwatters:
Looks like the license check is failing because of the `deno_runtime` dependency on the [`notify`](https://crates.io/crates/notify) crate which has the ` CC0-1.0 OR Artistic-2.0 ` licenses that aren't green listed.

I know if I have access to approve additional licenses but also wouldn't want to make a change without owner involvement.

auraescript/src/lib.rs
krisnova: I would be inclined to let the client user set and manage these permissions independently of the project. I would honestly just turn the conversation around and have AuraeScript very clearly (AKA "loudly") advertise (EG: In our docs, at the CLI, etc) the permissions it requires to run.

For example we will likely need more detail around this statement:

> access to APIs that provide information about user's operating system

Will AuraeScript be able to mutate kernel parameters on the client's machine? If so? Which ones? Which Linux Capabilities will be required to access a socket on localhost, and so forth.

If we can begin the conversation witih at least documenting these I am fine with it. We can drill into the specifics of each of these later.

krisnova (approved):
This is a great starting point, we can follow up more on the `AllowAll()` consequences in future PRs.

Thanks!

seanwatters:
After some trial and error, was able to figure out that `deno_runtime::MainWorker` requires usage of `force_op_registration` to get ops on the `Deno[Deno.internal].core` object, and that `initializeAsyncOps` has been removed and calling `async` host operations appears to only be possible with `Deno[Deno.internal].core.opAsync` now. 

I've put up a draft PR to address these issues: https://github.com/aurae-runtime/aurae/pull/474. When testing locally I was able to get all of the Deno errors resolved, but then have a new error and it isn't clear to me whether it's related to Deno changes or something else but appears to be unrelated.

```bash
warning: failed to parse config at /home/parallels/.aurae/config: TOML parse error at line 39, column 10
   |
39 | socket = "/var/run/aurae/aurae.sock"
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
unknown variant `/var/run/aurae/aurae.sock`, expected `Path` or `IPv6`

warning: failed to parse config at /etc/aurae/config: No such file or directory (os error 2)
warning: failed to parse config at /var/lib/aurae/config: No such file or directory (os error 2)
Error: Uncaught (in promise) Error: unable to find valid config file
```

dmah42:
that seems unrelated.. but I'm not sure why you'd see it.  have you run make pki and make config?

maybe we are parsing the config before the socket file has been created and someone changed the config parsing to be stricter.

PR #472, opened 2023-03-26
More efficient build/lint/test commands in Makefile

future-highway:
Changes based on the observations noted [here](https://github.com/aurae-runtime/aurae/issues/461#issuecomment-1478091020)

Essentially grouping commands on target. Also, only `auraed` is installed for the tests now, as only it is required to be installed for the tests.


dmah42 (approved):

future-highway:
001 looks like it will fail from timing out. Since 001 and 101 are the same with 101 just being in a container, I'll override and merge.

PR #474, opened 2023-03-27
fix(auraescript): removal of initializeAsyncOps and exposing Deno.core.ops for deno_runtime::MainWorker

seanwatters:
The `initializeAsyncOps` function has been deprecated and no longer available on latest `deno_runtime`/`deno_core`. Deno prefers usage of `Deno.core.opAsync("op_name", ...args)`.

Additionally, `deno_runtime::MainWorker` does not initialize ops unless you use the `force_op_registration` method on `ExtensionBuilder`.

Updates `Deno.core` to `Deno[Deno.internal].core` for `MainWorker` as it "hides" host operations with an internal namespace. 

Un-async-ifies `as__aurae_config__try_default`, `as__aurae_config__from_options` and `as__aurae_config__parse_from_file` which do not make use of the `await` keyword or any async paradigms.


auraescript/src/bin/main.rs
```diff
@@ -65,8 +65,8 @@ fn main() -> anyhow::Result<()> {
     let main_module = resolve_path(&args[1].clone(), current_dir()?.as_path())?;
     let mut main_worker = init(main_module.clone());
 
-    let _ =
-        main_worker.execute_script("", "Deno.core.initializeAsyncOps();")?;
+    let _ = main_worker
+        .execute_script("", "Deno.core = Deno[Deno.internal].core;")?;
```
dmah42, line 69: I assume deno put that into an internal namespace on purpose. is exposing it safe?

dmah42 (approved):
well done for figuring this out.  just one question but it looks reasonable to me.

auraescript/src/bin/main.rs
seanwatters: I believe so, I think they mostly wanted to add _some_ level of privacy, but it is still always technically accessible via `Deno[Deno.internal]`.

The reason why I sent it "up a level" was just to minimize the changes that needed to be made to existing TypeScript files and TypeScript codegen. But I could also see an argument for just using `Deno[Deno.internal]` in the TypeScript code because someone who is used to seeing regular Deno code might be confused and would have to find this code to understand why `Deno.core` works.

Open to either option.

dmah42: what you've done is fine, it's just a little odd that they (deno) decided to put expected public functionality into an internal namespace.

dmah42 (approved):

seanwatters:
@dmah42 this should be all good to go now. Am going to open another PR for the removal of `aurae.print` once this one is merged.

To test whether or not it's working I ran `cargo run ../examples/cells.ts` from the `auraescript` workspace and everything looked good.

seanwatters:
here is the PR for removing `aurae.print`: https://github.com/aurae-runtime/aurae/pull/477

PR #475, opened 2023-03-28
Fix parsing a socket path in config

future-highway:
By default, the toml crate does not handle enum (de)serialization. This PR implements the traits needed by serde to enable deserializing an `AuraeSocket`. Currently it assumes the config is specifying a `Path`, but I would think we also want to support specifying a `SocketAddrV6` at some point.

### DIFF
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -120,7 +120,10 @@ impl Client {
                     .await
             }
             AuraeSocket::IPv6 { ip: mut socket, scope_id } => {
-                socket.set_scope_id(scope_id);
+                if let Some(scope_id) = scope_id {
+                    socket.set_scope_id(scope_id);
+                }
+
                 endpoint
                     .connect_with_connector(service_fn(move |_: Uri| {
                         TcpStream::connect(socket)
--- a/client/src/client.rs
+++ b/client/src/client.rs
@@ -33,8 +33,6 @@
 //! Manages authenticating with remote Aurae instances, as well as searching
 //! the local filesystem for configuration and authentication material.
 
-use std::net::SocketAddrV6;
-
 use crate::config::{AuraeConfig, CertMaterial, ClientCertDetails};
 use crate::AuraeSocket;
 use thiserror::Error;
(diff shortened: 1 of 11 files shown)


client/src/client.rs
```diff
@@ -121,9 +119,7 @@ impl Client {
                     }))
                     .await
             }
-            AuraeSocket::IPv6 { ip, scope_id } => {
-                let mut socket =
-                    ip.parse::<SocketAddrV6>().expect("invalid ip address");
+            AuraeSocket::IPv6 { ip: mut socket, scope_id } => {
```
dmah42, line 122: so does this path not work any more?

future-highway: It should work, if you have a way to specify the IPv6. With this PR, path has been brought back as being able to be specified, but a SocketAddr has not (partly because I don't know what a `scope_id` is and how we would represent an ip + scope_id in toml).

future-highway: I'm not sure what PR needed the changes as I haven't had enough time lately to keep up with them all. It was definitely working, but I assume the change was needed for something.

According to [this](https://networkengineering.stackexchange.com/questions/46653/what-is-the-use-of-the-ipv6-scope-id) (and the RFC it links), `%` is the separator used, so I'll use that. The examples of the scope id aren't u32 though.

dmah42: https://manpages.ubuntu.com/manpages/trusty/man7/ipv6.7.html <-- u32

client/src/config/mod.rs
```diff
@@ -167,4 +169,25 @@ socket = "/var/run/aurae/aurae.sock"
...
+    #[test]
+    fn can_parse_toml_config_socket_ipv6() {
+        const INPUT: &str = r#"
+[auth]
+ca_crt = "~/.aurae/pki/ca.crt"
+client_crt = "~/.aurae/pki/_signed.client.nova.crt"
+client_key = "~/.aurae/pki/client.nova.key"
+
+[system]
+socket = "[fe80::2]:8080%4"
```
dmah42, line 182: nit: we should also test not having the scope id as it is optional.

dmah42 (dismissed):
thank you!

client/src/config/mod.rs
future-highway: In system_config.rs there is `can_parse_aurae_socket_ipv6` and `can_parse_aurae_socket_ipv6_with_scope_id` which should cover both scenarios. 

I was actually going to take this test out, but forgot :). I'm not sure it adds anything. over the path one, but if you see value in it I can add the not scope one here too.

PR #477, opened 2023-03-30
Removes `aurae.print` which is no longer necessary because `console.log` is available

seanwatters:


dmah42 (dismissed):

dmah42 (dismissed):

izissise:
I believe `toString` is only used in `print` maybe it can be removed aswell?

seanwatters:
Great point. Pushed up the `toString` removal.

dmah42 (approved):

PR #478, opened 2023-04-09
Pin youki crates to commit

future-highway:
Since youki is in early development, relying on its main branch has been causing issues where auraed ends up breaking unexpectedly as changes are made upstream. I would have expected our Cargo.lock to save us, but it hasn't, so these changes make the dependencies rely on the specific commit instead of branch.

Relying on a branch seems like it would be pretty unstable in general (vs a commit or tag). We also rely on a branch for aya in our ebpf crates. I wanted to change that in this PR as well, but I couldn't get the ebpf crates to build (likely due to me being on a aarch64 system and aya has changed the dependencies it needs, which I could not successfully install).


dmah42 (approved):

JeroenSoeters (approved):
Build is green again! :ok_hand:

PR #484, opened 2023-06-20
Bump deno deps to resolve udeps failure

dmah42:
The underlying type of `TypeId` changed from 64-bit to 128-bit and this
now causes issues in udeps unless we bump deno (which in turn bumps v8).


dmah42:
"error[E0432]: unresolved import `tokio_metrics::RuntimeMonitor`"

PR #485, opened 2023-06-30
get auraescript building

dmah42:


dmah42:
docker checks appear to be timing out, presumably due to the new set of dependencies.
udeps is passing locally so i'm not sure why that's failing unless it's also caching the set of dependencies somehow?

(i'm a bit grumpy about our docker based CI stuff.  i do wish we could just use bare metal and run the same commands that we expect our developers to locally.  ie, `make`).

future-highway (approved):
`make auraescript` and `make auraed` completed successfully for me locally. 

Unfortunately, my local environment is not able to run the tests (not a new issue).

dmah42:
> `make auraescript` and `make auraed` completed successfully for me locally. 
> 
> Unfortunately, my local environment is not able to run the tests (not a new issue). 

do you know why udeps is failing here but no locally? it's almost like it's using an old set of dependencies.

PR #487, opened 2023-08-07
Migrate from buf's remote generation alpha

dmah42:
more build issues... still hacking away

### DIFF
--- a/auraescript/src/cells.rs
+++ b/auraescript/src/cells.rs
@@ -28,4 +28,6 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
+use deno_core::Op;
+
 macros::ops_generator!("../api/v0/cells/cells.proto", cells, CellService,);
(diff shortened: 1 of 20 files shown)


dmah42:
> check deny and check docs are timeouts. so i think this is ready to go.

timeouts resolved by extending them. check deny is still freaking out but i'm not sure how big a deal that is :)

dmah42:
seems like somehow youki/libcgroups lost its LICENSE file between v0.0.4 and v0.0.5...

auraescript/src/cells.rs
```diff
@@ -28,4 +28,6 @@
  *                                                                            *
 \* -------------------------------------------------------------------------- */
 
+use deno_core::Op;
```
future-highway, line 31: Nit: preferably this would be generated by the macro

future-highway (approved):
I was able to build, install and run aurae and auraescript (`cells.ts` example) successfully.

Also looked at the check deny run and agree that the errors don't look like a big deal.

PR #489, opened 2023-11-15
fix clippy lint errors with latest cargo

dmah42:
updating the dependencies seems to help with the newer clippy lint errors.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.20.0"
+version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jesse-peters (approved):
Looks good to me

future-highway (approved):
Didn't do a local build, but LGTM. 

From my understanding, `resolver = "2"` is the default since we are only using edition 2021, but I assume you had a reason to add it explicitly. 

Edit: Just realized I put `resolver = "2"` in my own project too (can't remember why).

dmah42:
> Didn't do a local build, but LGTM. 
> 
> From my understanding, `resolver = "2"` is the default since we are only using edition 2021, but I assume you had a reason to add it explicitly. 
> 
> Edit: Just realized I put `resolver = "2"` in my own project too (can't remember why).

it was a cargo warning about a mismatch between the default (1) and the default for 2021 (2).

PR #492, opened 2024-01-14
Fix 404s on Rustdocs for `auraed`

mccormickt:
The below pages were 404s on the website.
* https://aurae.io/crate/implementors
* https://aurae.io/crate/macros
* https://aurae.io/crate/src/auraed/lib.rs.html

This PR removes the unused `/macros` and `/implementors` routes and unifies the rustdocs to fix the documentation. Includes some drive-by typo fixes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,7 +315,7 @@ endif
 
 .PHONY: docs-crates
 docs-crates: musl $(GEN_TS) $(GEN_RS) ## Build the crate (documentation)
-	$(cargo) doc --target $(uname_m)-unknown-linux-musl --no-deps --package auraed
+	$(cargo) doc --no-deps --package auraed
 	$(cargo) doc --no-deps --package auraescript
 	$(cargo) doc --no-deps --package client
 	$(cargo) doc --no-deps --package aer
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -315,7 +315,7 @@ endif
 
 .PHONY: docs-crates
 docs-crates: musl $(GEN_TS) $(GEN_RS) ## Build the crate (documentation)
-	$(cargo) doc --target $(uname_m)-unknown-linux-musl --no-deps --package auraed
```
mccormickt, line 318: Not sure if this breaks anything. Assuming our CI would run with the correct rust target by default, I think we're fine. This change allows all the rustdocs to be unified and have links to each crate on the left navigation.

image

dmah42: > Not sure if this breaks anything. Assuming our CI would run with the correct rust target by default, I think we're fine. This change allows all the rustdocs to be unified and have links to each crate on the left navigation.
> 
> image

it shouldn't, no. our ci doesn't use the right target (musl) but for docs I suspect it doesnt matter...

dmah42 (approved):

dmah42:
the unused import should be unrelated (and I'm not sure how it snuck in).  let me see if I can clean that up first.

PR #493, opened 2024-01-14
remove unused use of allocation::Allocation

dmah42:


mccormickt (approved):
I should have a fix for the cargo deny failures with #494

PR #491, opened 2024-01-12
chore: update dependencies

BradLugo:

PR #496, opened 2024-01-20
Upgrade Deno

mccormickt:
This PR implements the necessary changes that come with upgrading to the latest version of `deno_runtime`. 

* Utilizes new [`op2`](https://github.com/denoland/deno_core/tree/main/ops/op2) macro to define our auraescript functions.
* Utilizes runtime snapshots for quicker bootstrapping of the runtime ([details](https://deno.com/blog/roll-your-own-javascript-runtime-pt3#creating-a-snapshot-in-buildrs)).
* Adapts the use of removed `opAsync` function in Typescript to directly supported async functions.
* Also removes the direct dependency on `deno_core` in favor of the re-export from `deno_runtime` to help keep these crates aligned.

Should help unblock other dependency updates as well in #491


auraescript/build.rs
```diff
@@ -1,9 +1,16 @@
+use deno_runtime::snapshot;
 use std::fs::OpenOptions;
 use std::io::Write;
 use std::path::PathBuf;
 
 fn main() {
     generate_aurae_ts();
+
+    // Create a runtime snapshot for Deno to be used on startup
+    snapshot::create_runtime_snapshot(
```
mccormickt, line 10: Snapshots are now "required" in Deno without passing a couple feature flags to load javascript sources for bootstrapping. This decreases startup performance pretty noticeably, so I went with the snapshot approach. Some info in [this issue](https://github.com/denoland/deno/issues/21928).

```
# With snapshot
» time ./examples/discovery_health.ts 
[Object: null prototype] { healthy: true, version: "0.0.0" }
./examples/discovery_health.ts  0.05s user 0.00s system 75% cpu 0.067 total

# Without snapshot
» time ./examples/discovery_health.ts
[Object: null prototype] { healthy: true, version: "0.0.0" }
./examples/discovery_health.ts  1.82s user 0.02s system 99% cpu 1.847 total
```

mccormickt:
Cargo deny failed due to a BUSL license for a transitive, windows-only dep in `deno-runtime`. Limited cargo deny check targets to x86_64/aarch64 linux in c895aa6, but can totally move that commit to a separate PR if we deem it fit.

dmah42 (approved):

dmah42:
Tester failing due to space on device still.  i think we need to separate out the github actions to build each subproject.

PR #499, opened 2024-01-23
downgrade deno as deno_ast 1.0.1 was unpublished

dmah42:


MalteJ (approved):
tested. worked. merged.

PR #504, opened 2024-05-28
adding some basic documentation for auraed source code

taniwha3:
Signed-off-by: Tani Aura <111664369+taniwha3@users.noreply.github.com>

### DIFF
--- a/auraed/src/cells/cell_service/cell_service.rs
+++ b/auraed/src/cells/cell_service/cell_service.rs
@@ -60,22 +60,30 @@ use tokio::sync::Mutex;
 use tonic::{Code, Request, Response, Status};
 use tracing::{info, trace, warn};
 
+/**
+ * Macro to perform an operation within a cell.
+ * It retries the operation with an exponential backoff strategy in case of connection errors.
+ */
 macro_rules! do_in_cell {
     ($self:ident, $cell_name:ident, $function:ident, $request:ident) => {{
+        // Lock the cells mutex to ensure exclusive access
         let mut cells = $self.cells.lock().await;
 
+        // Retrieve the client socket for the specified cell
         let client_socket = cells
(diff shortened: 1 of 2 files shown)


auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -60,22 +60,30 @@ use tokio::sync::Mutex;
 use tonic::{Code, Request, Response, Status};
 use tracing::{info, trace, warn};
 
+/**
+ * Macro to perform an operation within a cell.
+ * It retries the operation with an exponential backoff strategy in case of connection errors.
+ */
 macro_rules! do_in_cell {
     ($self:ident, $cell_name:ident, $function:ident, $request:ident) => {{
+        // Lock the cells mutex to ensure exclusive access
```
dmah42, line 69: in general I'm not a huge fan of comments that describe the code that is already readable.  I prefer to keep comments for methods/classes and any tricky bits of code.  that way readers know to pay attention to the comments that exist and they don't become background noise that readers ignore.  does that make sense?

auraed/src/cells/cell_service/cell_service.rs
```diff
@@ -145,6 +170,13 @@ impl CellService {
         })
     }
 
+    /// Frees a cell. Will delocate the cell.
```
dmah42, line 173: "deallocate"?

dmah42 (approved):

GANitak:
FWIW I think the commentary lines can only **help understand the step better** for someone cruising the code without having a large amount of experience with this **kind of code, or language.**
So it helps me, yes some of the lines are obvious, but others are not.
Concerning the background noise, I can see that too.
 _Maybe there is a way to hide comment sections somehow...?_ 
But more accessible information directly from the code base, a gain of time for the people. 
Can comments affect runtime seriously?

PR #506, opened 2024-06-20
Implement VM Service with Cloud Hypervisor

mccormickt:
This PR builds off of the foundations by @JeroenSoeters in #451. Instead of requiring the cloud-hypervisor binary running externally and using the client, this implementation uses the `vmm` crate for CH to manage aurae `VirtualMachines`. We also leave configuring TAP devices up to Cloud Hypervisor instead of creating them out of band. This should allow auraed to create VMs given 

Using Cloud Hypervisor's VMM like this has a few nice properties:
* Auraed remains a free-standing executable.
* Security, maintainability, and compatibly benefits provided by the `rust-vmm` ecosystem and CH community.
* Requests are served directly with no additional marshalling/unmarshalling or connections required to serve an additional an API server (if we were to use CH's http API).
* Minimal config for networking, CH takes care of TAP creation, routing, and teardown

The tests implemented also allow you to run virtual machines with an Auraed as PID 1 after running the below copying the kernel and disk image to proper paths under `/var/lib/aurae`:
```
sudo make /opt/aurae/cloud-hypervisor/cloud-hypervisor
sudo make build-guest-kernel
sudo make prepare-image
```

A few future improvements not in this PR:
* PID 1 runtime enhancements to configure networking and mounting filesystems (e.g. cgroups to allow Cells to be created, etc.).
* Validation on `VirtualMachine` configs and status
* Routing and external connectivity for VMs
* Design how we handle VMs with/without nested Auraed's (e.g. how do we configure the socket address, poll for status of nested daemon, if it exists)


auraed/src/vms/vm_service.rs
```diff
@@ -1,45 +1,129 @@
...
+        }));
+
+        let net = vec![NetSpec {
+            tap: Some(format!(
+                "aurae0-{}",
+                rand::rand_alphanumerics(8).into_string().map_err(|_| {
+                    Status::internal("Failed to generate tap device name")
+                })?
+            )),
+            ip: Ipv4Addr::new(192, 168, 122, 1),
```
dmah42, line 62: TODO: Ipv6?
TODO: rotate the IP addresses on each call to create (if i'm reading this code right)?

auraed/tests/vms_start_must_start_vm_with_auraed.rs
```diff
@@ -1,61 +1,61 @@
...
+    );
+
+    assert!(res.is_ok(), "{:?}", res);
+
+    let res = retry!(
+        client.start(VmServiceStartRequest { vm_id: vm_id.clone() }).await
+    );
+
+    assert!(res.is_ok(), "{:?}", res);
+    thread::sleep(std::time::Duration::from_secs(10));
```
dmah42, line 47: it would be nice to not need this sleep.. is there a signal we can get (in future) from the health or discovery service  to poll?

mccormickt: Absolutely - I'll add an endpoint to list VMs and use that combined with the VMs status field to know when to attempt to build the client and connect to the VMs Auraed.

auraed/src/vms/virtual_machine.rs
```diff
@@ -0,0 +1,332 @@
...
+        if let Some(sender) = &manager.sender {
+            vmm::api::VmCreate
+                .send(
+                    manager.events.try_clone()?,
+                    sender.clone(),
+                    Arc::new(Mutex::new(spec.clone().into())),
+                )
+                .expect("Failed to send create request");
+        } else {
+            return Err(anyhow!("Virtual machine manager not initialized"));
```
mccormickt, line 203: Should add custom error types for the VM module.

auraed/src/vms/virtual_machines.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+impl VirtualMachines {
+    /// Create a new instance of the virtual machines cache.
+    pub fn new() -> Self {
+        unsafe {
+            let _ = libc::signal(libc::SIGCHLD, libc::SIG_IGN);
+        }
+
+        // Before we start any threads, mask the signals we'll be
```
mccormickt, line 23: This is adapted almost directly from Cloud Hypervisor ([here](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/98d0aabb99dcdbf5e8a1e6ebc0aa1ab72b3c1114/src/main.rs#L595-L613)). I'm not quite sure whether this needs to be run before each VM is provisioned, or just at startup.

auraed/src/vms/vm_service.rs
mccormickt: >rotate the IP addresses on each call to create (if i'm reading this code right)?
Yes, definitely should do that. I could start by implementing it naively and just iterate through a hard-coded subnet?

>Ipv6?
We can access over IPv6 link-local (`fe80::`) already, but it doesn't look like CH supports assigning ipv6 addresses to interfaces yet. I think we'd need set up the TAP manually to do this.

https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/net_util/src/open_tap.rs#L91

auraed/src/vms/virtual_machine.rs
dmah42: yes please! though a TODO is fine for now and someone can pick it up as a quick win.

auraed/src/vms/vm_service.rs
dmah42: > >rotate the IP addresses on each call to create (if i'm reading this code right)?
> 
> Yes, definitely should do that. I could start by implementing it naively and just iterate through a hard-coded subnet?

oh yeah that's fine.  254 VMs should be enough for anyone!

> 
> >Ipv6?
> 
> We can access over IPv6 link-local (`fe80::`) already, but it doesn't look like CH supports assigning ipv6 addresses to interfaces yet. I think we'd need set up the TAP manually to do this.

let's add a TODO.  I'd very much like to be ipv6 only by default all through the stack.  I'm a bit surprised CH hasn't made the same decision.

but it's all internal so doesn't matter that much.

> 
> https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/net_util/src/open_tap.rs#L91

auraed/tests/vms_start_must_start_vm_with_auraed.rs
mccormickt: Took a stab at something in [f08afa7](https://github.com/aurae-runtime/aurae/pull/506/commits/f08afa77ab8738229d7daa80291be07d581dca2c). 

I attempted to address both polling the VM's status and fetching the nested `auraed` socket address to facilitate connecting to it.

api/v0/vms/vms.proto
```diff
@@ -46,25 +46,58 @@ service VmService {
...
+  // The number of vCPUs for the VM
+  uint32 vcpu_count = 4;
+
+  // The path to the VM kernel image
+  string kernel_img_path = 5;
+
+  // Path to the image mounted as the root directory of the VM
+  string root_dir_path = 6;
+
+  // Auraed server address of the VM
```
dmah42, line 78: does this need a  port or is it host:port?  ipv4 or ipv6?

auraed/src/init/system_runtimes/pid1_system_runtime.rs
```diff
@@ -104,6 +108,8 @@ impl SystemRuntime for Pid1SystemRuntime {
         let socket_addr = socket_address
             .unwrap_or_else(|| DEFAULT_NETWORK_SOCKET_ADDR.into())
             .parse::<SocketAddr>()?;
-        create_tcp_socket_stream(socket_addr).await
+        let stream = create_tcp_socket_stream(socket_addr).await?;
```
dmah42, line 111: isn't this the same?  `await` and `?` then wrap in `Ok` seems the same as just returning the results of the `await` directly.

auraed/src/vms/virtual_machine.rs
```diff
@@ -0,0 +1,382 @@
...
+            tpm: None,
+            preserved_fds: None,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct NetSpec {
+    pub tap: Option<String>,
+    pub ip: Ipv4Addr,
```
dmah42, line 108: can we do Ipv6 by default? or does that give us some compatibility issues?

auraed/src/vms/virtual_machines.rs
```diff
@@ -0,0 +1,129 @@
...
+        // installing handlers for, to make sure they only ever run on the
+        // dedicated signal handling thread we'll start in a bit.
+        for sig in &vmm::vm::Vm::HANDLED_SIGNALS {
+            if let Err(e) = block_signal(*sig) {
+                eprintln!("Error blocking signals: {e}");
```
dmah42, line 35: should these be fatal errors?  we should also integrate with the logging we have elsewhere (use tracing::{info, trace, warn};)

auraed/tests/common/mod.rs
```diff
@@ -104,9 +105,7 @@ async fn run_auraed() -> Client {
             e => break e,
         }
     }
-    .expect("failed to create client");
-
-    client
```
dmah42, line 109: was this unreachable before?

auraescript/src/vms.rs
```diff
@@ -0,0 +1,33 @@
...
+ *                |  ███████║██║   ██║██████╔╝███████║█████╗   |              *
+ *                |  ██╔══██║██║   ██║██╔══██╗██╔══██║██╔══╝   |              *
+ *                |  ██║  ██║╚██████╔╝██║  ██║██║  ██║███████╗ |              *
+ *                |  ╚═╝  ╚═╝ ╚═════╝ ╚═╝  ╚═╝╚═╝  ╚═╝╚══════╝ |              *
+ *                +--------------------------------------------+              *
+ *                                                                            *
+ *                         Distributed Systems Runtime                        *
+ *                                                                            *
+ * -------------------------------------------------------------------------- *
+ *                                                                            *
```
dmah42, line 16: check the new shorter header format :)

auraed/src/init/system_runtimes/pid1_system_runtime.rs
mccormickt: Removed in https://github.com/aurae-runtime/aurae/pull/506/commits/9a35b2fc94984777d84b99156fff80194f434b05

auraed/src/vms/virtual_machine.rs
mccormickt: I'm just letting Cloud Hypervisor set up the networking here, which seems to automatically set an IPv4 address by as well as an IPv6 link-local by default. I think we would just need to set up the tap device manually to allow this to be configurable.

```
tap0: flags=67<UP,BROADCAST,RUNNING>  mtu 1500
        inet 192.168.249.1  netmask 255.255.255.0  broadcast 192.168.249.255
        inet6 fe80::68af:9eff:fed6:8823  prefixlen 64  scopeid 0x20<link>
        ether 06:66:42:a8:3f:e1  txqueuelen 1000  (Ethernet)
        RX packets 9  bytes 822 (822.0 B)
        RX errors 0  dropped 0  overruns 0  frame 0
        TX packets 21  bytes 2182 (2.1 KB)
        TX errors 0  dropped 0 overruns 0  carrier 0  collisions 0
```

dmah42: fair enough. thanks for confirming

auraed/src/vms/virtual_machines.rs
mccormickt: Swapped for tracing in https://github.com/aurae-runtime/aurae/pull/506/commits/1d289d24a98567afff84f634376b020f228b8d4b

> should these be fatal errors?

I'm honestly not sure. The only thing I gather is that the VMM "should" be handling these signals explicitly, but CH [doesn't seem to care](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/9f9a4e657f96cc292ac34354292377a275752819/src/main.rs#L636-L646) if something goes wrong.

auraed/tests/common/mod.rs
mccormickt: This was fixing a clippy lint, "returning the result of a `let` binding from a block"

auraescript/src/vms.rs
mccormickt: Amazing - updated https://github.com/aurae-runtime/aurae/pull/506/commits/ed1eaa6757bde628617d5bf7aaec9d3508cefba7

api/v0/vms/vms.proto
mccormickt: This is currently a link-local ipv6 host:port which is allocated regardless of whether there actually is an auraed set to run on the VM.

It may be best to potentially remove all "nested" auraed assumptions in this PR and just get an initial baseline of "creating VMs" before we come together and decide out how we want to handle auraed in VMs.

dmah42 (approved):

PR #513, opened 2024-06-23
chore(auraescript): Upgrade Deno

mccormickt:
Fixes some build errors due to dependency conflicts by upgrading Deno libraries. We also split `deno_runtime` and `deno_core` again to allow for macros to work properly (re-exported `deno_core` could not be found). 

Also disables the snake case clippy lint for auraescript modules as it now fails on the `as__` prefixed functions generated by our macros.

### DIFF
--- a/auraescript/src/lib.rs
+++ b/auraescript/src/lib.rs
@@ -65,19 +65,14 @@
 #![allow(unused_qualifications)]
 
 use anyhow::{anyhow, bail, Error};
-use deno_ast::{EmitOptions, MediaType, ParseParams, SourceTextInfo};
+use deno_ast::{EmitOptions, MediaType, ParseParams, SourceMapOption};
 use deno_core::{
-    self, error::AnyError, futures::FutureExt, resolve_import, url::Url,
+    self, error::AnyError, resolve_import, url::Url, JsRuntime,
     ModuleLoadResponse, ModuleLoader, ModuleSource, ModuleSourceCode,
     ModuleSpecifier, ModuleType, RequestedModuleType, ResolutionKind,
+    RuntimeOptions, SourceMapGetter,
 };
-use deno_runtime::{
-    permissions::PermissionsContainer,
-    worker::{MainWorker, WorkerOptions},
-    BootstrapOptions, WorkerLogLevel,
-};
-
-use std::rc::Rc;
+use std::{cell::RefCell, collections::HashMap, future::Future, rc::Rc};
 
 mod builtin;
 mod cells;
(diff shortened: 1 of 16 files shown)


mccormickt:
Seems like this breaks `auraescript`, converting to draft to figure it out. The usual issue of the ops not being registered properly.

```
» ./examples/cells.ts 
Error: TypeError: Deno[Deno.internal].core.ops.as__aurae_config__try_default is not a function
    at Module.createClient (file:///home/jan0ski/aurae-runtime/aurae/auraescript/gen/aurae.ts:15:47)
    at file:///home/jan0ski/aurae-runtime/aurae/examples/cells.ts:4:28
```

mccormickt:
Seems to be working well! A few errors that seem to be more like bugs in `auraed` not cleaning up cgroups/executables or in the scripts themselves rather than in `auraescript`:

```
» ./examples/discovery_discover.ts
{
  "healthy": true,
  "version": "0.0.0"
}
```

```
» ./examples/cells_isolated_processes_and_network.ts 
{
  "cellName": "ae-net-proc-isolated-1",
  "cgroupV2": true
}
{
  "pid": 19
}
{
  "pid": 21
}
{
  "pid": 25
}
Error: Error: status: Internal, message: "cell 'ae-net-proc-isolated-1' could not kill children: No child process (os error 10)", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Wed, 26 Jun 2024 01:41:53 GMT", "content-length": "0"} }
    at async file:///home/jan0ski/aurae-runtime/aurae/examples/cells_isolated_processes_and_network.ts:84:1


```

dmah42:
interesting.  do we see this before the upgrade?

mccormickt:
>interesting. do we see this before the upgrade?

After a restart these "no process found" or "cgroup exists" errors are gone (both on main and this branch). The problem seemed to be stale cgroups on my machine from prior tests not being cleaned up 👍

dmah42:
> >interesting. do we see this before the upgrade?
> 
> After a restart these "no process found" or "cgroup exists" errors are gone (both on main and this branch). The problem seemed to be stale cgroups on my machine from prior tests not being cleaned up 👍 

so is this ready to merge?

auraescript/src/lib.rs
```diff
@@ -164,12 +166,15 @@ impl ModuleLoader for TypescriptModuleLoader {
...
+        _requested_module_type: RequestedModuleType,
+    ) -> ModuleLoadResponse {
+        fn load(
+            source_maps: SourceMapStore,
+            module_specifier: &ModuleSpecifier,
+        ) -> Result<ModuleSource, Error> {
             let path = module_specifier
                 .to_file_path()
-                .map_err(|_| anyhow!("Only file: URLs are supported."))?;
+                .map_err(|_| anyhow!("Only file:// URLs are supported."))?;
```
dmah42, line 177: should we use the requested module type instead of extracting it from the path?  or maybe use the requested module path and the path-derived version of None is passed in?

dmah42 (approved):

PR #518, opened 2024-06-28
simplify make actions

dmah42:


dmah42:
pr check is, ironically, failing.  i'll need to figure out which tests are a problem first before submitting this as i want a cleaner Makefile but not at the expense of an all-in-one PR check workflow.

dmah42:
do we need `make prcheck`?

the Tester actions today run `make <project>-test` which runs `cargo test` on the project (and builds/lints auraed)

`prcheck` builds and lints all the projects and runs test-all, which includes "ignored" tests.

maybe what we need is:

1. Tester workflows that run `-lint` and `-test` per project
2. non-required `-test-all` workflow that is maintained as best-effort.

wdyt?

dmah42:
looks like some youki/dependency issue in the 10x builders.

dmah42:
"thread 'cells_list_must_list_allocated_cells_recursively' panicked at auraed/tests/common/mod.rs:93:6:
failed to create client: Other(Failed to read server root CA certificate from path '/etc/aurae/pki/ca.crt'"

looks solvable, but maybe not as part of this PR?

mccormickt:
> looks solvable, but maybe not as part of this PR?

I think we can get this to pass and not add too much to the other make targets by just adding the `pki` target as a dependency for the `pr-check` command? I think this could be optional as well if the ignored tests become a problem.

https://github.com/aurae-runtime/aurae/pull/518/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R100-R101

Otherwise, the make workflows that I personally use work great.

PR #520, opened 2024-07-22
Upgrading dependencies, hopefully making the build green again

MalteJ:


auraed/Cargo.toml
```diff
@@ -52,26 +52,26 @@ clone3 = "0.2.3"
 fancy-regex = { workspace = true }
 futures = "0.3.28"
 ipnetwork = "0.20.0"
-iter_tools = "0.1.4"
+iter_tools = "0.20.0"
 libc = "0.2.155" # TODO: Nix comes with libc, can we rely on that?
 lazy_static = { workspace = true }
-libcgroups = { git = "https://github.com/containers/youki", rev = "f4e7e300e6be7a4ea758a436218b9db1b39c69cd", default-features = false, features = [
+libcgroups = { git = "https://github.com/containers/youki", rev = "5b62356e377def45c36c29183c586c4302685cf8", default-features = false, features = [
```
dmah42, line 58: can we get to a tag?

dmah42 (approved):
thank you! I've had this half finished for a couple of weeks.

auraed/Cargo.toml
MalteJ: No. I tried the latest tag 0.3.3 but it uses the struct `oci_spec::runtime::LinuxRlimit` which exist in oci-spec v0.6.4 but has been renamed to `oci_spec::runtime::PosixRlimit` in v0.6.8. Unfortunately the oci_spec crate is referenced as `~0.6.4` in the libcontainer crate (https://github.com/containers/youki/blob/c03edea8a0655c97baad3216fba46a47531ac5f0/crates/libcontainer/Cargo.toml#L42).

PR #523, opened 2024-08-04
Remove container-based gha workflows

bpmooch:
Ran into some issues on my other pr, after discussion, this pr seemed prudent and hopefully helps simplify debugging for me

This pr also contains some ancillary commits to quiet the linter temporarily until some tech debt is resolved


bpmooch:
ill squash this if checks pass

Makefile
```diff
@@ -334,48 +336,10 @@ else
 	sudo -E ./hack/serve.sh
 endif
 
-#------------------------------------------------------------------------------#
-
-# Container Commands
```
dmah42, line 339: I thought these were used for general development rather than for CI.  oe.we should still have some of these for building local containers and images to run against.  is that not right?

README.md
```diff
@@ -1,183 +1,203 @@
-<!-- THE DOCUMENT -->
+# protoc-gen-doc
```
dmah42, line 1: this change seems wrong.

Makefile
bpmooch: I have no idea. What are the current local use cases people have for the containers?

dmah42: chatted in discord: convenience for managing OCI images for those that want to use that workflow, and as a form of distribution using docker.  i think we only need the busybox/alpine targets for manual testing .. the others can all  go afaict.

dmah42 (dismissed):

dmah42 (approved):

dmah42:
that branch protection rule can't pass any more as you removed the CI builds.

PR #525, opened 2024-08-09
Add hack/install-os-deps.sh

bpmooch:
Added a script to install os dependencies for building aurae and used it in a github workflow to test it out

### DIFF
--- a/.github/workflows/001-tester-ubuntu-make-test-auraed.yml
+++ b/.github/workflows/001-tester-ubuntu-make-test-auraed.yml
@@ -30,7 +30,7 @@
 #
 # When this build passes we should have a "clean cargo" output!
 #
-name: "Tester (001) [ubuntu:latest] [auraed]"
+name: "(001) [ubuntu:latest] make auraed-test"
 on:
   push:
     branches: main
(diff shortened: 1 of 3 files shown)


.github/workflows/001-tester-ubuntu-make-test-auraed.yml
```diff
@@ -63,38 +63,9 @@ jobs:
...
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y musl-tools
       - name: Cargo build, lint, test [make test]
         # This should remain the only command we execute as this matches the title of the file.
         # The goal is for this to be easy to find from the GitHub dashboard.
         # Instead of adding more commands here, consider another make target or a new YAML file
         # named with a good name.
-        run: make auraed-test
+        run: make os-deps auraed-test
```
dmah42, line 71: should auraed-test just depends on os-deps?  or on auraed-build which depends on os-deps... 

that would make things easier for people getting started I think.

bpmooch: hrm I'm not sure what the best ux is. Most likely `os-deps` needs to get added to one or more of the super targets, but not sure which. Thoughts?

bpmooch:
if checks pass, I might update 002 and 003 gh workflows with the same refactor I did to 001

.github/workflows/001-tester-ubuntu-make-test-auraed.yml
dmah42: leave this as it is (and yes please update the other workflows) and I'll see what I can do.

dmah42 (approved):

PR #526, opened 2024-08-09
Update workflows to use install-build-deps.sh

bpmooch:
Taking what worked from 001 and spreading it to 002 and 003, also updated some random license headers, obviously automation on those needs a tweak


dmah42 (dismissed):

Makefile
```diff
@@ -88,16 +73,16 @@ clean: clean-certs clean-gens clean-crates ## Clean the repo
 lint: auraed-lint not-auraed-lint ## Run all lints
 
 .PHONY: test
-test: auraed-build auraed-lint auraed-test not-auraed-build not-auraed-lint not-auraed-test ## Builds, lints, and tests (does not include ignored tests)
+test: os-deps auraed-build auraed-lint auraed-test not-auraed-build not-auraed-lint not-auraed-test ## Builds, lints, and tests (does not include ignored tests)
```
dmah42, line 76: probably not the right place to put it.  we want it where we check for buf so if but it's missing we can install it (same for the proto gen stuff) 

remove this and just keep this pr focused on the workflow changes to call it and I can clean all the dependency stuff up later.

dmah42 (changes requested):

Makefile
bpmooch: heard

dmah42 (approved):

.github/workflows/001-tester-ubuntu-make-test-auraed.yml
```diff
@@ -63,7 +58,7 @@ jobs:
         with:
           toolchain: stable
           components: clippy, rustfmt
```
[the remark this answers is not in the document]
bpmooch: i think @0xE282B0 is right, we'll see

bpmooch:
@dmah42 sorry this hung around so long, should be good to land now

dmah42 (approved):

dmah42:
thank you so much

PR #528, opened 2024-08-13
Only insert executable into cache once it starts

dmah42:
The observability hooks are created in `Executable::new` so we shouldn't miss any messages.  This ensures that we only add started executables to the cache, which means that we don't end up with orphan executables in the cache.


future-highway (approved):
Looks good.

There is a comment about changing `contains_key` to `try_insert` in the future, which seems like it could accidently revert this change, so maybe remove that comment too?

future-highway:
I don't know how to unapprove...give me a sec...might be a bug

PR #532, opened 2024-08-16
Introduce uid and gid to the CellService start method

dmah42:
This allows the caller to override the uid and gid of the spawned executable within the Cell.  The response now includes the uid and gid, even if they've been inherited from the parent auraed.


auraed/src/cells/cell_service/executables/executable.rs
```diff
@@ -65,17 +64,27 @@ impl Executable {
...
+        uid: Option<u32>,
+        gid: Option<u32>,
+    ) -> io::Result<()> {
         let ExecutableState::Init { command } = &mut self.state else {
             return Ok(());
         };
 
-        let mut child = command
+        let mut command = command
             .kill_on_drop(true)
```
mccormickt, line 77: Is this variable dropped after this function exits? I'm curious if this setting would cause the cell process to be killed almost immediately, and whether it should be removed.

mccormickt (approved):

auraed/src/cells/cell_service/executables/executable.rs
dmah42: the functionality should be analogous to what happens today.

my understanding is that because we store the child in the `ExecutableState` it isn't dropped.  is that understanding incorrect?

mccormickt: Ah yes I missed where we saved the child to our state. 👍🏼

Issue #537, opened 2025-02-10
Compile Aurae on ARM


0xE282B0:
When compiling `auraed` on a `aarch64` machine, I get a compiler error that there are missing fields in `vmm::config::VmConfig`. These fields are [only available for `x86_64` targets](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/vmm/src/vm_config.rs#L540-L558).

```bash
error[E0560]: struct `vmm::config::VmConfig` has no field named `debug_console`
   --> auraed/src/vms/virtual_machine.rs:134:13
    |
134 |             debug_console: DebugConsoleConfig::default(),
    |             ^^^^^^^^^^^^^ `vmm::config::VmConfig` does not have this field
    |
    = note: all struct fields are already assigned

error[E0560]: struct `vmm::config::VmConfig` has no field named `sgx_epc`
   --> auraed/src/vms/virtual_machine.rs:141:13
    |
141 |             sgx_epc: None,
    |             ^^^^^^^ `vmm::config::VmConfig` does not have this field
    |
    = note: all struct fields are already assigned
```

PR #538, opened 2025-02-10
fix(auraed,vmm): make x86_64 only fields conditional

0xE282B0:
This PR makes `auraed` compile `aarch64` by making the fields that [only apply to `x86_64`](https://github.com/cloud-hypervisor/cloud-hypervisor/blob/main/vmm/src/vm_config.rs#L540-L558) dependent on the architecture.

Closes https://github.com/aurae-runtime/aurae/issues/537

### DIFF
--- a/auraed/src/vms/virtual_machine.rs
+++ b/auraed/src/vms/virtual_machine.rs
@@ -12,6 +12,7 @@
  * Copyright 2022 - 2024, the aurae contributors                              *
  * SPDX-License-Identifier: Apache-2.0                                        *
 \* -------------------------------------------------------------------------- */
+use crate::vms::manager::Manager;
 use anyhow::anyhow;
 use net_util::MacAddr;
 use std::{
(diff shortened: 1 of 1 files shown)


auraed/src/vms/virtual_machine.rs
```diff
@@ -98,13 +102,15 @@ impl From<VmSpec> for vmm::vm_config::VmConfig {
             pmem: None,
             serial: default_serial(),
             console: default_console(),
+            #[cfg(target_arch = "x86_64")]
             debug_console: DebugConsoleConfig::default(),
```
dmah42, line 106: it makes me grumpy that we need to essentially copy the same constraints as are in upstream, because this means a future maintenance headache as fields are added or constraints are changed (what happens when upstream supports arm? how do we know to include it now?) but i guess it's Rust WAI.

dmah42 (dismissed):

dmah42 (approved):

PR #539, opened 2025-02-12
chore(auraed): update auraed dependencies

0xE282B0:
This PR updates almost all dependencies of auraed to the latest version, only `rtnetlink` and `netlink-packet-route` remain from the last version before a breaking change, as I'm not sure if the necessary changes in the PID 1 code would break anything.

- [x] Build & Test runns on aarch64
- [x] Build & Test runns on x86_64


dmah42 (approved):

dmah42:
thank you!

PR #540, opened 2025-02-13
chore(ebpf): update aya to 0.13.1 and adapt code

0xE282B0:
This PR updates aya to 0.13.1 which is the latest version, it had some minor breaking changes mostly due to renaming.

- [x] Builds on aarch64
- [x] Builds on x86_64

The `ebpf/` subproject is not built in our workflows. Should we add a `make ebpf` workflow so that we can at least see that the code compiles?


auraed/src/ebpf/perf_buffer_reader.rs
```diff
@@ -32,11 +32,19 @@ const PER_CPU_BUFFER_SIZE_IN_PAGES: usize = 2;
...
-        let num_cpus = nr_cpus()?;
+        let num_cpus = nr_cpus().map_err(|(path, error)| {
+            std::io::Error::new(
+                error.kind(),
+                format!(
+                    "Failed to get the number of CPUs at {}: {}",
+                    path, error
+                ),
+            )
+        })?;
```
0xE282B0, line 47: I'm sure there is a more elegant solution.
`nr_cpus()` now returns `Result<size, (&'static str, io::Error)>`.
A quick fix would be `.map_err(|(_, error)| error)` but we'd lose the path. Idk if the path is important information here.

mccormickt:
> The `ebpf/` subproject is not built in our workflows. Should we add a `make ebpf` workflow so that we can at least see that the code compiles?

From what I recall, Github actions restricted us from installing eBPF programs in their runners. Due to this, we have those tests disabled in CI.  I'm not sure if this has changed, but we could surely revisit.

dmah42 (dismissed):

auraed/src/ebpf/perf_buffer_reader.rs
dmah42: > I'm sure there is a more elegant solution.
> `nr_cpus()` now returns `Result<size, (&'static str, io::Error)>`.
> A quick fix would be `.map_err(|(_, error)| error)` but we'd lose the path. Idk if the path is important information here.

I can't think of one either. an odd decision by Aya where an io::Error could be constructed with the path in, or a custom Error could be used...  but while I don't think the path is that important this approach is fine.

dmah42 (approved):

dmah42:
> This PR updates aya to 0.13.1 which is the latest version, it had some minor breaking changes mostly due to renaming.
> 
> - [x] Builds on aarch64
> - [x] Builds on x86_64
> 
> The `ebpf/` subproject is not built in our workflows. Should we add a `make ebpf` workflow so that we can at least see that the code compiles?

that's a good idea, even if we can't run it.

PR #548, opened 2025-03-05
fix(flake): cleanup `flake.nix`

ph:
Clean the `flake.nix`:
- Follow the _unstable_ channel for NixOS to get buf version 1.50.0.
- Remove unecessary shell hooks.
- Remove rustup from the packages list.
- Get the rust version from the `rust-toolchain.toml` file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,5 @@ auraescript/node_modules*
 pki*
 **/gen/
 hypervisor/guest-kernel/linux-cloud-hypervisor
+/.direnv/
+/.envrc
--- a/flake.lock
+++ b/flake.lock
@@ -1,12 +1,15 @@
 {
   "nodes": {
     "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
       "locked": {
-        "lastModified": 1659877975,
-        "narHash": "sha256-zllb8aq3YO3h8B/U0/J1WBgAL8EX5yWf5pMj3G0NAmc=",
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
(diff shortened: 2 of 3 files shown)


0xE282B0:
Thanks, this looks great.
I'm not a nix person, but I was told that the Determinate Nix Installer and `nix develop` would be the way to use it. I think it looks like a good alternative to the scripts we use to make sure the right versions of tools are installed.

However, while `make install auraed-start` works fine, `make ebpf` fails because it [checks if a nightly toolchain is available and installs it if needed](https://github.com/aurae-runtime/aurae/blob/main/ebpf/Makefile#L44-L47).
Is it possible to install multiple toolchains like with rustup?

Since I had no idea how to use `nix` and others might have the same problem, you might be interested in promoting nix a bit by describing [how to use it for aurae development in the docs](https://github.com/aurae-runtime/aurae/tree/main/docs/development-environments). 😇

ph:
I haven't run the 'ebf' target I did run the default one. I will take a look, but yes you can have more than one iirc.

dmah42 (approved):

0xE282B0:
The eBPF target is broken and needs to be updated, but we have a PR for that.

dmah42:
fwiw i'd also support not having the nix stuff in the repo if we had to choose.  originally there was going to be some broader embracing of nix and nixos but it didn't end up happening, and we don't really have the expertise.

ph:
@dmah42 I think it's still good to have it the repository, it's really a pain to have it outside of the repo. It can be done but it's really not a pleasant experience. I am more inclined to think if there are nix files and a contributor is using them, they will fix it and keep them up to date.

That was the main reason for me, but i might be wrong :)

dmah42:
that's fair. my only concern is maintenance of multiple systems but I'm happy for passionate folks to contribute.

PR #549, opened 2025-03-05
chore(buf): Fix buf warning

ph:
Replace `DEFAULT` with `STANDARD` as recommended by the warning message when running `buf lint api`.

Message excerpt:
```
Category DEFAULT referenced in your buf.yaml is deprecated. It has been replaced by category STANDARD.

[...]

As with all buf changes, this change is backwards-compatible: DEFAULT will continue to work.
We recommend replacing DEFAULT in your buf.yaml, but no action is immediately necessary.
```

### DIFF
--- a/api/buf.yaml
+++ b/api/buf.yaml
@@ -4,7 +4,7 @@ breaking:
     - FILE
 lint:
   use:
-    - DEFAULT
+    - STANDARD
   ignore:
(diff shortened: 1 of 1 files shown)


0xE282B0:
LGTM.

dmah42 (approved):

PR #555, opened 2025-06-03
fix(auraed): restore musl rust target

### LINKED ISSUE #554: Auraed should build with musl by default
In #524 we removed the make targets for installing the `musl` linux target in lieu of a `.cargo/config.toml`.

The [`targets`](https://rust-lang.github.io/rustup/overrides.html#targets) section is only additive, and does not override the default cargo target:
```
» rustup default
stable-x86_64-unknown-linux-gnu (default)
» make auraed-build                                                                                     130 ↵
cargo build -p auraed --locked
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.23s
» file ./target/debug/auraed
./target/debug/auraed: ELF 64-bit LSB pie executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, BuildID[sha1]=e80feca9ddcfa6c168fd6d6fe09b3af999807d43, for GNU/Linux 3.2.0, with debug_info, not stripped
```

Additionally, this will break being able to run as pid 1 in a vm, as it fails to find linked objects:
```
» sudo make prepare-image
» sudo cloud-hypervisor --kernel /var/lib/aurae/vm/kernel/vmlinux.bin \
--disk path=/var/lib/aurae/vm/image/disk.raw \
--cmdline "console=hvc0 root=/dev/vda1 rw" \
--cpus boot=4 \
--memory size=4096M \
--net "tap=tap0,mac=aa:ae:00:00:00:03,id=eth0"

[..snip..]

[    0.123590] Run /sbin/init as init process
/sbin/init: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by /sbin/init)
/sbin/init: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /sbin/init)
/sbin/init: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /sbin/init)
/sbin/init: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /sbin/init)
[    0.127868] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100
[    0.129692] Kernel Offset: disabled
[    0.129740] ---[ end Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000100 ]---
```

mccormickt:
Restores the `--target` flags to the makefile build targets.

Fixes #554


dmah42 (approved):

dmah42:
> `error: 'cargo-clippy' is not installed for the toolchain '1.85.0-x86_64-unknown-linux-gnu'.`
> 
> i think we have to be more sophisticated about how it's added.

just needed a merge

PR #558, opened 2025-06-04
Add signal for graceful shutdown of VMs

mccormickt:
Figured we should have parity here between Cells and VMs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -185,7 +185,7 @@ define AURAE_template =
 $(1): $(GEN_RS) $(GEN_TS) $(1)-lint $(1)-debug ## Lint and install $(1) (for use during development)
 
 .PHONY: $(1)-lint
-$(1)-lint: $(GEN_RS) $(GEN_TS)
+$(1)-lint: fmt $(GEN_RS) $(GEN_TS)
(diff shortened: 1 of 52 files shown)


auraed/src/graceful_shutdown.rs
```diff
@@ -77,15 +86,25 @@ impl GracefulShutdown {
         // wait for all subscribers to drop
         self.shutdown_broadcaster.closed().await;
 
+        if let Err(e) = self.cell_service.stop_all().await {
```
mccormickt, line 89: This seemed out of order. We should try and stop before freeing, I'd imagine?

dmah42: good catch.

auraed/src/graceful_shutdown.rs
```diff
@@ -77,15 +86,25 @@ impl GracefulShutdown {
...
+            error!("Attempt to stop all executables on terminate resulted in error: {e}")
```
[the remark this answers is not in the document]
mccormickt: Done in https://github.com/aurae-runtime/aurae/pull/558/commits/590b178d1b76e4d926ad3e5bf2f1e353e47fda18

dmah42 (approved):

PR #593, opened 2026-01-18
refactor(observe): make Arc internal to proc and cgroup caches

mccormickt:
This PR moves `Arc`'s and/or `Mutex`'s to be internal to `LogChannel`, `ProcCache` and `CgroupCache`. This cleans up a lot of instances where we must instantiate wrap the instantiation of the structs in an `Arc<T>` or `Arc<Mutex<T>>` and where we must handle the mutex and locking when accessing the struct's data.


dmah42 (dismissed):

dmah42:
looks great. one test failed though.

mccormickt:
> looks great. one test failed though.

Thanks! Yeah accidentally flipped the order of operations in `stop()`. Looks good now.

dmah42 (approved):

PR #601, opened 2026-05-06
refactor(ebpf): use `AsyncFd` in `PerfEventReader`

mccormickt:
`AsyncPerfEventArray` and the accompanying `async_tokio` is removed in future versions of `aya`. This change refactors our `PerfBufferReader` to use an `AsyncFd` for async support to reduce friction in migrating to a future release.

https://github.com/aya-rs/aya/pull/1292

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -383,7 +383,6 @@ dependencies = [
  "object",
  "once_cell",
  "thiserror 1.0.69",
- "tokio",
 ]
 
 [[package]]
--- a/auraed/Cargo.toml
+++ b/auraed/Cargo.toml
@@ -28,7 +28,7 @@ path = "src/bin/main.rs"
 anyhow = { workspace = true }
 client = { workspace = true }
 aurae-ebpf-shared = { path = "../ebpf-shared" }
-aya = { version = "0.13.1", features = ["async_tokio"] }
+aya = { version = "0.13.1" }
 backoff = { version = "0.4.0", features = ["tokio"] }
 bytes = "1.2.1"
 clap = { workspace = true }
--- a/auraed/src/ebpf/perf_buffer_reader.rs
+++ b/auraed/src/ebpf/perf_buffer_reader.rs
@@ -16,13 +16,16 @@
 use anyhow::Context;
 use aya::{
     Ebpf,
-    maps::perf::AsyncPerfEventArray,
+    maps::perf::PerfEventArray,
     util::{nr_cpus, online_cpus},
 };
 use bytes::BytesMut;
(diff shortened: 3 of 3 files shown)


ebpf/src/probe-kprobe-taskstats-exit.rs
```diff
@@ -12,26 +12,6 @@
  * Copyright 2022 - 2024, the aurae contributors                              *
  * SPDX-License-Identifier: Apache-2.0                                        *
```
mccormickt, line 13: Maybe we need to adapt this header for the code since it must be GPL-compatible? Cillium (Apache 2.0) has their ebpf code licensed [Dual BSD/GPL](https://github.com/cilium/cilium/pull/18858), for example.

dmah42: it looks like your change here is removing the GPL-2.0 and MIT license identifiers that were here for exactly that reason. (it was dual license MIT/GPL2.0 prior to this PR)

ebpf/src/probe-tracepoint-sched-sched-process-fork.rs
```diff
@@ -12,26 +12,6 @@
  * Copyright 2022 - 2024, the aurae contributors                              *
  * SPDX-License-Identifier: Apache-2.0                                        *
 \* -------------------------------------------------------------------------- */
-/* -------------------------------------------------------------------------- *\
```
dmah42, line 15: same here.. don't remove this :)

ebpf/src/probe-tracepoint-signal-signal-generate.rs
```diff
@@ -12,26 +12,6 @@
  * Copyright 2022 - 2024, the aurae contributors                              *
  * SPDX-License-Identifier: Apache-2.0                                        *
 \* -------------------------------------------------------------------------- */
-/* -------------------------------------------------------------------------- *\
```
dmah42, line 15: ditto

ebpf/src/probe-tracepoint-sched-sched-process-fork.rs
mccormickt: Gotcha - thought this was just a double-header mistake.

ebpf/src/probe-kprobe-taskstats-exit.rs
mccormickt: Reverted. Limited this change to just the tokio-compat and left the headers and licenses as they were.

dmah42 (approved):

PR #602, opened 2026-08-19
refactor(cells): async cell allocate/free

mccormickt:
Replace the `proxy_if_needed!` macro and make `CellsCache` async. `broadcast_kill` stays the only sync function, as it is called by `Drop`. Likewise, `do_free!` is replaced by `teardown_process_and_cgroup`.

This allows cells to free concurrently, so a host with many cells tears down in roughly the time of the slowest cell rather than the sum of all of them. This becomes important once we're tearing down network devices, etc.

### DIFF
--- a/auraed/src/cells/cell_service/cells/cell.rs
+++ b/auraed/src/cells/cell_service/cells/cell.rs
@@ -18,44 +18,12 @@ use super::{
     nested_auraed::NestedAuraed,
 };
 use client::AuraeSocket;
+use std::io;
 use tracing::info;
 
 // TODO https://github.com/aurae-runtime/aurae/issues/199 &&
 //      aurae.io/signals, which is more accurate
 // TODO nested auraed should proxy (bus) POSIX signals to child executables
-
-macro_rules! do_free {
-    (
-        $self:ident,
-        $nested_auraed_call:ident($($nested_auraed_call_arg:ident),*),
-        $($children_call:ident($($children_call_arg:ident),*)),*
-    ) => {{
-        if let CellState::Allocated { cgroup, nested_auraed, children } =
-            &mut $self.state
-        {
-            $(children.$children_call($($children_call_arg),*));*;
-
-            let _exit_status = nested_auraed
-                .$nested_auraed_call($($nested_auraed_call_arg),*)
-                .map_err(|e| {
-                    CellsError::FailedToKillCellChildren {
-                        cell_name: $self.cell_name.clone(),
-                        source: e,
-                    }
-                })?;
-
-            cgroup.delete().map_err(|e| CellsError::FailedToFreeCell {
-                cell_name: $self.cell_name.clone(),
-                source: e,
-            })?;
-        }
-
-        // set cell state to freed, independent of the current state
-        $self.state = CellState::Freed;
-
-        Ok(())
-    }};
-}
 
 // We should not be able to change a cell after it has been created.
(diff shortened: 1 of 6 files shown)


auraed/src/cells/cell_service/cells/cell.rs
```diff
@@ -141,15 +129,39 @@ impl Cell {
...
+        Ok(())
     }
 
     /// Sends a [SIGKILL] to the [NestedAuraed], and deletes the underlying cgroup.
     /// The [Cell::state] will be set to [CellState::Freed] regardless of it's state prior to this call.
     /// A [Cell] should never be reused once in the [CellState::Freed] state.
+    ///
+    /// Stays synchronous so [`Drop`] can call it.
     pub fn kill(&mut self) -> Result<()> {
-        do_free!(self, kill(), broadcast_kill())
```
dmah42, line 152: do_free! was a useful common utility shared between free and kill. is there a benefit to removing it?

auraed/src/cells/cell_service/cells/cells.rs
```diff
@@ -61,77 +38,134 @@ impl Cells {
...
+        }
 
-            let cell = self
-                .cache
-                .entry(cell_name.clone())
-                .or_insert_with(|| Cell::new(cell_name, cell_spec));
+        // From here, we know the cgroup doesn't exist, so remove from cache
+        // if it does
+        if let Some(_removed) = self.cache.remove(&cell_name) {
+            // TODO: Should we not remove the cell (that has no cgroup) from
```
dmah42, line 86: while you're here, can you think about whether this TODO makes sense? maybe we can implement it if you think it does or remove the comment if you think it doesn't.

dmah42 (approved):

auraed/src/cells/cell_service/cells/cells.rs
mccormickt: I was going through the scenarios in my head and sheepishly left this unresolved. Thinking out loud:

* We have exclusive control over auraed-managed cgroups due to the mutex over `Cells` so no threat of a race.
* Since we prefix the names, I don't think cgroup name collision is possible for child auraeds.
* Something may have failed when it allocated it in the first place (or something else cleaned up the cgroup), so should we just re-allocate it?
* Similarly on line 100 - If we remove from cache on allocate failure, then the above case should only happen if something outside auraed removed the cgroup.

Maybe removing on failure to allocate + overwriting existing cell without a cgroup is sensible?

auraed/src/cells/cell_service/cells/cell.rs
mccormickt: Maybe a bit premature.  This was a small change I was attempting to peel off as a prerequisite for a few more features I've been working on, like cell networking (issue / write-up coming!). 

Once we're managing network devices, `Cell::free` will need to be async to make netlink calls. `do_free!` put the differing functions (sync/async) inside the expansion so I refactored the same logic into a function accepting a closure instead to satisfy this.

auraed/src/cells/cell_service/cells/cells.rs
dmah42: i'm a fan of explicit failure modes for unexpected events, but that can be quite destructive _if_ things outside auraed are monkeying with our cgroups. we can leave this as a comment for now until we discuss or think more.

dmah42 (approved):

